PR #19, opened 2018-04-24
Add 3D ROI and fix slicing and zoom issues

rsmith013:


paskino:
Let's merge #20 first and then update this one.

PR #28, opened 2018-05-10
Added ability to load multiple tiff images for visualisation

rsmith013:


paskino:
you've already merged this, but I guess this script should not go into the `site-packages` rather into the `bin` (not the rubbish one), so that the user with the package could run it by calling it in a terminal.

rsmith013:
yes, perhaps I was a little hasty in pushing this through. I was thinking about creating a little script to run this and then placing the launch to it on the start menu.

PR #30, opened 2018-05-14
QT GUI updates

rsmith013:
- Constrained QT viewer to downsample data if supplied more than a 512 cube.
- Added entry point to meta.yml so that QT app can be run from the command line. Can be used to run viewer in VM from start menu.


paskino:
Please (squash and) merge it

PR #64, opened 2020-03-25
Updated parseNpyHeader() so that it successfully reads .npy files.

lauramurgatroyd:
This is needed so that we can view .npy image files in vtk.

### DIFF
--- a/Wrappers/Python/ccpi/viewer/utils/conversion.py
+++ b/Wrappers/Python/ccpi/viewer/utils/conversion.py
@@ -951,11 +951,12 @@ def parseNpyHeader(filename):
         # print ("header_len", HEADER_LEN, type(HEADER_LEN))
         descr = ''
         i = 0
-    with open(filename, 'r') as f:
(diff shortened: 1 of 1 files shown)


Wrappers/Python/ccpi/viewer/utils/conversion.py
```diff
@@ -951,11 +951,12 @@ def parseNpyHeader(filename):
         # print ("header_len", HEADER_LEN, type(HEADER_LEN))
         descr = ''
         i = 0
-    with open(filename, 'r') as f:
+    with open(filename, 'rb') as f:
         f.seek(6+2 + HEADER_LEN_SIZE)
 
         while i < HEADER_LEN:
             c = f.read(1)
+            c = c.decode("utf-8")
```
paskino: This could also be a one liner
```python
c = f.read(1).decode("utf-8")
```
I suppose. 

I wonder why `open(filename, 'r', encoding='utf-8')` didn't work.

Issue #94, opened 2020-09-11
Preserve image spacing and origin when resampling and cropping images


lauramurgatroyd:
Currently the BaseResampleReader and BaseCroppedReader in conversion.py do not preserve the original origin and spacing of the image file.

In the resample reader the new image spacing needs to be the downsampled spacing multiplied by the original image spacing. The new origin needs to be the original origin -0.5 to center the slices correctly.

paskino:
I suppose the origin should be `-0.5 * slice_thickness`, i.e. `1/spacing`.

lauramurgatroyd:
Since our discussion yesterday the above comment re: the origin is now incorrect. The new origin should be: 

origin_resampled = (1 - spacing) * 0.5 + origin_orig

as discussed here: https://github.com/vais-ral/CCPi-DVC/issues/109

PR #97, opened 2020-09-21
Fix coordinate systems in the viewer

lauramurgatroyd:
Fixes:

- [x]  Conversion of image and world coordinates - including correct offset due to origin
- [x]  Display of coordinate annotation
- [x]  Preserves original image origin, and makes use of original image spacing when downsampling and cropping


paskino (changes requested):
Apparently, the way the viewer calculates the world coordinate is wrong ([here](https://github.com/vais-ral/CILViewer/blob/92d4092d2c07e3fb2d8b8f6e9e32a47b8e5a5066/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L901), [here ](https://github.com/vais-ral/CILViewer/blob/92d4092d2c07e3fb2d8b8f6e9e32a47b8e5a5066/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L887)and [here](https://github.com/vais-ral/CILViewer/blob/92d4092d2c07e3fb2d8b8f6e9e32a47b8e5a5066/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L873)) because it scales the origin with the spacing, while (I believe) the origin should be indipendent.

```python

world_coord = image_coord * spacing - origin
# so that
origin_resampled = (1 - spacing) * 0.5 + origin_orig
```

lauramurgatroyd:
The corner annotation needs updating to use the image2world method

The following lines: https://github.com/vais-ral/CILViewer/blob/92d4092d2c07e3fb2d8b8f6e9e32a47b8e5a5066/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L1694

We can delete lines:
1694-1698
1704-1709
1714-1719

And add at 1691:

``` python
if self.display_unsampled_coords and self.image_is_downsampled:
    data = image2world(data)
    
```

paskino (approved):
I presume it's ok! As it takes care of the new spacing!

paskino:
I will have to rethink my approval!

lauramurgatroyd:
This PR needs updating with the correct definition of the world and image coordinates in the viewer.

What we need to do is as follows:
 

1.  **Fix the definition of the image and world coords in the viewer:** https://github.com/vais-ral/CILViewer/issues/98#issuecomment-697362561

2. **Adjust this PR as follows:**

Lines 808-815 should be replaced with:
 
```
new_spacing = [element_spacing[0]/xy_axes_magnification,
element_spacing[1]/xy_axes_magnification, 
element_spacing[2]/z_axis_magnification]
 
new_origin = [(i-1)/2 for i in new_spacing]
resampled_image.SetOrigin(new_origin[0], new_origin[1], new_origin[2])
```
 

An explanation for the origin in the old coordinate system is here: https://github.com/vais-ral/CCPi-DVC/issues/109#issuecomment-696402033, but I will re-explain with the new coordinate system here:
 
If we have an image which is downsampled by 5 times, slices 0-4 are downsampled to a single slice and the image spacing is 5. 
 
Slice 0 in image coordinates corresponds to slices 0-4 in the actual image. The clipping planes will need to include points ranging from -0.5 to 4.49. Therefore the slice needs to be centred half way through this range: at 2.
 
Because `world coordinates = image coords * spacing + origin`, we need the origin to be 2 for this image.
 
In general, the origin must be at `(image_spacing-1)/2 `

This will close: https://github.com/vais-ral/CILViewer/issues/99

 
**3.	Update the corner annotation in the viewer** (currently it is poorly written)

Here: https://github.com/vais-ral/CILViewer/blob/92d4092d2c07e3fb2d8b8f6e9e32a47b8e5a5066/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L1694
We can delete lines:
1693-1698
1704-1709
1714-1719

And add at 1691:

``` python
if self.display_unsampled_coords and self.image_is_downsampled:
    data = image2world(data)
    
```

This will mean the "Downsampled Image Coords" shows the image coords, and the "Original Image Coords" shows the world coords.

**4. Update the Resample readers so that they preserve the original image's origin if it exists.**

For example, here: https://github.com/vais-ral/CILViewer/blob/369600bd4a86873b930e1bdadd52c7268a7936b8/Wrappers/Python/ccpi/viewer/utils/conversion.py#L745

it assumes the original image origin was (0,0,0).
This also needs to be changed for the classes that inherit from this one.

Then once we have done this, we can modify step 2 so that the origin of a downsampled image would be given by:
`(image_spacing-1)/2 + original origin `

This will close: https://github.com/vais-ral/CILViewer/issues/94

lauramurgatroyd:
The definition of image and world coords should now be correct, and displayed correctly in the corner annotation.

The downsampling and cropping image readers now take into account the original image origin, and spacing, which are read from a metaimage file. For the case of numpy files, I haven't seen an example where the origin or spacing is set in the numpy header (our numpy image writers don't have options to set these either).

paskino:
Please merge master in here and then I think we can have a last look at the PR together.

PR #111, opened 2021-05-04
HDF5 Read & Write

lauramurgatroyd:
- Allows reading of HDF5 files to vtkImageData with `HDF5Reader`, and reading a cropped version of the image with `RequestSubset`
- Allows writing of vtkImageData to HDF5 files with `write_image_data_to_hdf5`
- Resampling of HDF5 with `cilHDF5ImageResampleReader`
- Tests of the above

To do:

- [x]  Test RequestSubset works when selecting a channel
- [x]  Possibly rename methods
- [x]  Add resample reader


Wrappers/Python/ccpi/viewer/utils/conversion.py
```diff
@@ -946,29 +979,149 @@ def SetFileName(self, value):
     def GetFileName(self):
         return self.__FileName
 
-        
+
+class cilHDF5ImageResampleReader(cilBaseResampleReader):
```
lauramurgatroyd, line 983: The only change to conversion.py that isn't just formatting is the addition of this `cilHDF5ImageResampleReader` class

lauramurgatroyd:
Note:
I had to place the hdf5 methods into a new file hdf5_io. This was necessary to avoid circular imports.
The problem was that the HDF5Resample reader lives with the other resample readers in conversion.py, and this class requires the use of HDF5Reader. Also the ImageDataCreator in io requires the ResampleReaders. So the hdf5 classes couldn't be placed in io.

Wrappers/Python/ccpi/viewer/utils/hdf5_io.py
```diff
@@ -0,0 +1,184 @@
...
+    '''
+
+    def __init__(self):
+        VTKPythonAlgorithmBase.__init__(self,
+                                        nInputPorts=0,
+                                        nOutputPorts=1,
+                                        outputType='vtkImageData')
+
+        self.__FileName = ""
+        self.__Label = "ImageData"
```
paskino, line 60: is this reader reading NeXus as exported by CIL only?

lauramurgatroyd: No, it's not, you have to set the label where the data lives, maybe it's best that I change the default to:
`'entry1/tomo_entry/data/data'`

lauramurgatroyd: set to None by default and renamed to self.__DatasetName

lauramurgatroyd:
See https://github.com/vais-ral/CILViewer/pull/127 for improvements to structure of the resample reader

lauramurgatroyd:
Once #150 is working and merge, can check that the HDF5 test runs with the gh-action, and then will merge this

PR #112, opened 2021-05-08
Fix vtk9

paskino:
Removes import that made vtk9 fail.
Removes restriction to vtk 8.1.2 in meta.yaml


lauramurgatroyd (changes requested):
The update to vtk9 causes the tracer in the vtk app to stop working (vtkImageTracerWidget). We need to fix this before we can update the vtk version

lauramurgatroyd:
The issue with the tracer seems to arise from the fact that the[ vtkInteractorStyleImage ](https://vtk.org/doc/nightly/html/classvtkInteractorStyleImage.html#details) has the control: 

> Left Mouse button triggers window level events

Setting:
 https://github.com/vais-ral/CILViewer/blob/62c59fdb6677ce59403dbd63fe2197e76ff7ab2d/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L1352
Should mean that when we press t, the left mouse click results in tracing on the image. This happens in vtk8.1
But in vtk 9 this isn't the case.
If we change this line here: https://github.com/vais-ral/CILViewer/blob/62c59fdb6677ce59403dbd63fe2197e76ff7ab2d/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L1541  to use the output from self.voi ( vtk.vtkExtractVOI()) instead of self.wl ( vtk.vtkImageMapToWindowLevelColors()).
In this case the tracer works in vtk9.
Obviously we need to be able to change and set the window level. So we need an actual fix for this, but now we know where the problem may lie.
 
I also have a simplified example that demonstrates this. 
In this example, with vtk9 and input from the vtk.vtkImageMapToWindowLevelColors(), the tracer never works, even when t is pressed. Left click+drag always adjusts the window level.
When running the same with vtk8.1, when t is pressed, the trace works when you left click+drag, and when t is pressed again to deactivate trace mode instead of tracing it adjusts the window level when you left click+drag.

paskino:
Isn't there a way to consume the event `KeyPress` 't'? 

I'm thinking to enable the tracer programmatically, without relying on the widget to interpret the `KeyPress`

lauramurgatroyd:
The problem seems to be interpreting the left click to trace, rather than in getting the key press.
It gets the "t" keypress, because you can see the little cross appear and disappear, so the widget is activating and deactivating.

lauramurgatroyd:
The [vtkImageTracer ](https://vtk.org/doc/nightly/html/classvtkImageTracerWidget.html#details) should interpret interactions with the left mouse button, right mouse button and middle mouse button. 

Interestingly, the right mouse button interaction (clicking the x and dragging to move it) works, but the left and middle buttons don't.

lauramurgatroyd:
I tried out some things (none of which helped) which I'm going to comment on in a review
 
The example code that demonstrates how the vtk.vtkImageMapToWindowLevelColors() is causing problems, even outside of the viewer, (as in my above comment) is here: https://github.com/lauramurgatroyd/CCPi-Work/blob/main/CILViewer/vtk9.py
 
I even tried saving the output from the vtk.vtkImageMapToWindowLevelColors() to a metaimage file and then tried visualising that and the tracer still doesn't work.

Wrappers/Python/ccpi/viewer/CILViewer2D.py
```diff
@@ -76,10 +76,10 @@ def isAllInactive(self):
         return all(not x for x in self.events.values())
 
 
-class CILInteractorStyle(vtk.vtkInteractorStyleImage):
+class CILInteractorStyle(vtk.vtkInteractorStyle): #vtk.vtkInteractorStyleImage
```
lauramurgatroyd, line 79: Here I changed the interactor style we inherit from, in case it was the key press interactions from vtk.vtkInteractorStyleImage that was causing problems. Changing this made no difference.

Wrappers/Python/ccpi/viewer/CILViewer2D.py
```diff
@@ -96,6 +96,8 @@ def __init__(self, callback):
         self.AddObserver('LeftButtonReleaseEvent', self.OnLeftButtonReleaseEvent, priority)
         self.AddObserver('RightButtonReleaseEvent', self.OnRightButtonReleaseEvent, priority)
         self.AddObserver('MouseMoveEvent', self.OnMouseMoveEvent, priority)
+        self.AddObserver('CharEvent', self.OnCharEvent, priority)
```
lauramurgatroyd, line 99: Here I added an observer for CharEvent. Whenever we press a key eg. "t", it triggers both a CharEvent and a KeyPressEvent

Wrappers/Python/ccpi/viewer/CILViewer2D.py
```diff
@@ -1353,14 +1369,12 @@ def __init__(self, dimx=600,dimy=600, ren=None, renWin=None,iren=None, debug=Tru
         self.imageTracer.GetGlyphSource().SetRotationAngle(45.0)
         self.imageTracer.GetGlyphSource().Modified()
         self.imageTracer.ProjectToPlaneOn()
-        # Set key press activation on
-        self.imageTracer.KeyPressActivationOn()
-        # Use 't' to activate image tracer
-        self.imageTracer.SetKeyPressActivationValue('t')
+        self.imageTracer.SetHandleLeftMouseButton(True)
```
lauramurgatroyd, line 1372: This should be true by default anyway, but I tried setting it anyway

Wrappers/Python/ccpi/viewer/CILViewer2D.py
```diff
@@ -1353,14 +1369,12 @@ def __init__(self, dimx=600,dimy=600, ren=None, renWin=None,iren=None, debug=Tru
...
         
         # Set autoclose to on
         self.imageTracer.AutoCloseOn()
 
+        self.imageTracer.AddObserver(vtk.vtkWidgetEvent.Select, self.style.OnTracerModifiedEvent, 1.0)
```
lauramurgatroyd, line 1377: Added an observer for selecting the image tracer.
On left mouse press and middle mouse button, nothing happens
On right key press and hold we get `Tracer modified InteractionEvent` (this is what we should be getting with left and middle mouse too)
On "t" key press we get either:
Tracer modified DisableEvent
Tracer modified ModifiedEvent
or
Tracer modified ModifiedEvent
Tracer modified EnableEvent

Wrappers/Python/ccpi/viewer/CILViewer2D.py
```diff
@@ -545,7 +547,10 @@ def OnKeyPress(self, interactor, event):
             self.SetEventActive('UPDATE_WINDOW_LEVEL_UNDER_CURSOR')
         elif interactor.GetKeyCode() == "t":
             # tracing event is captured by widget
-            pass
+            if (self._viewer.imageTracer.GetEnabled()):
```
lauramurgatroyd, line 550: Changed it so we enable and disable the tracer ourselves by interpreting the key press, instead of the tracer doing this

paskino:
We use `vtkImageMapToWindowLevelColor` to change the window and level of the[ `vtkImageActor`](https://vtk.org/doc/nightly/html/classvtkImageActor.html#details) that we display. 

I read at the link that 
> [vtkImageActor] class has been mostly superseded by the vtkImageSlice class, which provides more functionality than vtkImageActor.

In particular in a [`vtkImageSlice` ](https://vtk.org/doc/nightly/html/classvtkImageSlice.html#details)
> The lookup table and window/level are set in `vtkImageProperty`. 

So if we used `vtkImageSlice` instead of `vtkImageActor` we would be able to not use `vtkImageMapToWindowLevelColor`.

Can you give it a go?

lauramurgatroyd:
Okay, I will try!

lauramurgatroyd:
It works with the ImageSlicer in [this](https://github.com/lauramurgatroyd/CCPi-Work/blob/main/CILViewer/vtk9_imageslice.py) example, so now I will add it into the viewer! 😃

paskino:
At any rate, one thing that we should learn to do is to pass the interaction we want to some widget.

For instance, I've got a similar problem with the rectilinear wipe widget in the viewer. The viewer interactor seems not to forward all the interaction, so it does not work (well).

I'm pretty sure we should consume all the events we capture and pass them to other widgets if we want.

lauramurgatroyd:
So far everything is working with the vtk.vtkImageSlice in the Z direction, I just haven't figured out how to change orientation yet

paskino:
It's the voi that determines the slice direction, no?

https://github.com/lauramurgatroyd/CCPi-Work/blob/76d2aca3ee6dc2e179fa6aff6fd3a5db84d9027e/CILViewer/vtk9_imageslice.py#L36-L63

I think we'll have a shorter pipeline with the output of the voi->imageslice mapper -> renderer going into the image slice mapper, and the imageslice added to the renderer. Avoiding the detour to the voi->window/level filter->mapper->actor -> renderer

Or am I missing something?

lauramurgatroyd:
I think that's what I did on this branch?

paskino:
Probably! The best is then to try on your minimal example.

paskino:
**Maybe** this tells you something about the slice orientation https://discourse.vtk.org/t/no-image-when-slicing-along-yz-or-xz-plane/6695

Maybe not, because we move the camera, don't we?

PR #116, opened 2021-05-17
Triggers action on pushing any tags

lauramurgatroyd:
Previously, action was not triggered when a tag was pushed


lauramurgatroyd:
Temporarily changing branch to current branch to test if publishing on pushing a tag works or not

lauramurgatroyd:
Just going to test that pushing more to this branch doesn't cause an upload incorrectly

lauramurgatroyd:
Releasing on tagging now correctly works, and we have a windows version published to conda.

paskino (approved):

Issue #104, opened 2021-04-01
Allow switching of orientation on the y axis


lauramurgatroyd:
To do this, we need to change the ViewUp vector in the vtkCamera

lauramurgatroyd:
We want to allow the user to not only change the y axis, but be able to specify which orientation we view each plane.

To do this, we need to:

- [ ] Set orientation along each axis explicitly
- [ ] Create setters for azimuth, elevation, view up vector, flipping camera postion etc.
- [ ] Create interface for selecting these options, probably with a viewer that only displays the orientation axis marker.

Currently the orientation of each axis is obtained by rotating from the current position. We need to change this so that the orientation on each axis is obtained independently of which axis you were previosuly on. Otherwise it will be very difficult to change the axes based on user input I think. Some (very broken) work in progress is on this branch: https://github.com/vais-ral/CILViewer/tree/change_axes

PR #118, opened 2021-05-24
Rotate orientation of axes 

lauramurgatroyd:
Rotates axes as follows:
image

### DIFF
--- a/Wrappers/Python/ccpi/viewer/CILViewer2D.py
+++ b/Wrappers/Python/ccpi/viewer/CILViewer2D.py
@@ -540,9 +540,9 @@ class CILInteractorStyle(vtk.vtkInteractorStyleImage):
             self.DisplayHelp()
         elif interactor.GetKeyCode() == "w":
             x,y = interactor.GetEventPosition()
-            print (x,y)
+            # print (x,y)
             ic = self.display2imageCoordinate((x,y))
-            print (ic)
+            # print (ic)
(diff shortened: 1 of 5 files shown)


Wrappers/Python/ccpi/viewer/CILViewer2D.py
```diff
@@ -1555,6 +1553,9 @@ def installPipeline(self):
         self.ren.ResetCamera()
         self.ren.Render()
 
+        self.camera.SetViewUp(0,-1,0)
+        self.camera.Azimuth(180)
```
paskino, line 1557: I suppose this should go in `AdjustCamera`?

paskino (approved):
I think this is fine, however in the future I'd like to be able to implement a more generic way to change the camera position and `viewUp` so that user can change it.

lauramurgatroyd:
I agree that we need to allow the user to set the orientation on each axis.
I started attempting it but managed to break a lot of things so started again and made this PR instead.
I added details of what I would like to do here: https://github.com/vais-ral/CILViewer/issues/104

Wrappers/Python/ccpi/viewer/CILViewer2D.py
lauramurgatroyd: This only happens in `installPipeline`, to make sure the initial camera direction is correct.
Whereas `AdjustCamera` is called in `updatePipeline` as well, where we don't want to move the camera to the xy plane

PR #137, opened 2021-08-20
Update to version 21.1.0

lauramurgatroyd:


lauramurgatroyd:
Not needed because https://github.com/vais-ral/CILViewer/pull/138 will deal with version updates automatically.

PR #138, opened 2021-08-20
create version from repo tag

paskino:
Infers the version from repository tag. 
Adds a way to get the version string programmatically by 
```python

from ccpi.viewer import version as vversion

print (vversion.version)

>>> v20.04-80-geaa5247
```

### DIFF
--- a/Wrappers/Python/setup.py
+++ b/Wrappers/Python/setup.py
@@ -16,12 +16,26 @@
 from distutils.core import setup
 import os
 import sys
+import subprocess
 
-cil_version = "21.0.0"
-if os.environ.get('CONDA_BUILD', None) is not None:
+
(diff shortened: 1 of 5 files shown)


lauramurgatroyd (approved):

Wrappers/Python/setup.py
```diff
@@ -16,12 +16,26 @@
...
 else:
     requires = ['numpy','vtk']
+    cwd = os.path.join(os.environ.get('RECIPE_DIR'),'..')
+
+fname = os.path.join(cwd, 'ccpi', 'viewer', 'version.py')
+
+if os.path.exists(fname):
+    os.remove(fname)
+with open(fname, 'w') as f:
+    f.write('version = \'{}\''.format(cil_version))
```
lauramurgatroyd, line 37: Writes out the variable as 'version' but then our import lines have 'dversion' so it breaks

PR #148, opened 2021-09-28
Set view up for y direction to avoid warning message

### LINKED ISSUE #145: Setting view to Y issues a warning about the view-up
Pressing Y we get this warning, meaning we are doing something not quite right.

```

Warning: In ..\Rendering\Core\vtkRenderer.cxx, line 1126
vtkOpenGLRenderer (000001ABA378CE00): Resetting view-up since view plane normal is parallel

Warning: In ..\Rendering\Core\vtkRenderer.cxx, line 1126
vtkOpenGLRenderer (000001ABA378C7A0): Resetting view-up since view plane normal is parallel

```

lauramurgatroyd:
Closes #145

### DIFF
--- a/Wrappers/Python/ccpi/viewer/CILViewer2D.py
+++ b/Wrappers/Python/ccpi/viewer/CILViewer2D.py
@@ -443,7 +443,8 @@ class CILInteractorStyle(vtk.vtkInteractorStyleImage):
             elif orientation == SLICE_ORIENTATION_XZ:
                 self.FlipCameraPosition(True)
                 camera.Azimuth(90)
-                camera.SetViewUp(0, -1, 0)
+            camera.SetViewUp(0, -1, 0)
+
             self.SetActiveCamera(camera)
             self.SetSliceOrientation(SLICE_ORIENTATION_YZ)
             self.UpdatePipeline(True)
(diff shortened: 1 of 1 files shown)


paskino (commented):
Why only for direction `x` we `SetViewUp` differently depending on the orientation before `x` was pressed?

https://github.com/vais-ral/CILViewer/blob/818f8db87cdc0f9a20249846360410cf47879bed/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L440-L446

lauramurgatroyd:
This is because in both the z direction and the x direction the view up vector is the same - see https://github.com/vais-ral/CILViewer/pull/118 - in both cases y points downwards.

This means also that here:
https://github.com/vais-ral/CILViewer/blob/818f8db87cdc0f9a20249846360410cf47879bed/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L490-L497
we could move the view up setting only if `orientation == SLICE_ORIENTATION_XZ` .

Or we could leave both outside to be more consistent and maybe less confusing

paskino (approved):

paskino:
I'd try to make this less confusing

PR #159, opened 2021-10-28
Noarch package

### LINKED ISSUE #157: package is noarch
I think we should set
```yaml

build:
  noarch: python
```

paskino:
defines package as noarch, remove convert to osx and win.
closes #157


lauramurgatroyd (commented):
Looks fine. We should also update this file: https://github.com/vais-ral/CILViewer/blob/master/.github/workflows/README.md

PR #164, opened 2021-11-10
removed io.py file

### LINKED ISSUE #162: io.py should be removed and/or handled differently
I guess the iDVC app is currently using the functionality in `utils/io.py` but we should remove this file.

paskino:
addresses removing file in #162


lauramurgatroyd:
I think it would be better for us to release a version for current master before doing this

PR #184, opened 2022-01-17
Add ErrorObserver and EndObserver for handling errors in our readers

### LINKED ISSUE #180: HDF5 reader errors when dataset name not set
Although this is expected, it fails in the wrong place.

https://github.com/vais-ral/CILViewer/blob/1e8ed88c42ddb1d6fd5f7834c60a917baf8580fc/Wrappers/Python/ccpi/viewer/utils/conversion.py#L1062-L1064

Here it should fail if the dataset name is set to `None`

lauramurgatroyd:
Also raises exceptions if essential inputs to readers are not set.
Example of using the ErrorObserver and EndObserver are in examples/error_observer.py

Closes #180

### DIFF
--- a/Wrappers/Python/ccpi/viewer/utils/conversion.py
+++ b/Wrappers/Python/ccpi/viewer/utils/conversion.py
@@ -754,11 +754,14 @@ class cilBaseResampleReader(VTKPythonAlgorithmBase):
         return self.__IsAcquisitionData
 
     def ReadDataSetInfo(self):
-        '''Not implemented for Base class
-        Tries to read info about dataset from given file
+        '''Tries to read info about dataset
         Will raise specific errors if inputs required for 
         file type are not set.'''
-        pass
+        if self.__StoredArrayShape is None:
+            raise Exception("StoredArrayShape must be set.")
+        
+        if self.__OutputVTKType is None:
+            raise Exception("Typecode must be set.")
 
     def _GetInternalChunkReader(self):
         tmpdir = tempfile.mkdtemp()
--- a/Wrappers/Python/ccpi/viewer/utils/conversion.py
+++ b/Wrappers/Python/ccpi/viewer/utils/conversion.py
@@ -600,7 +600,7 @@ class cilBaseResampleReader(VTKPythonAlgorithmBase):
     def __init__(self):
         VTKPythonAlgorithmBase.__init__(self, nInputPorts=0, nOutputPorts=1)
 
-        self.__FileName = 1
+        self.__FileName = None
         self.__TargetSize = 256*256*256
         self.__IsFortran = False
         self.__BigEndian = False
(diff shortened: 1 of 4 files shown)


lauramurgatroyd:
Temporarily converting to draft

Wrappers/Python/ccpi/viewer/utils/conversion.py
```diff
@@ -753,6 +753,13 @@ def SetIsAcquisitionData(self, value):
     def GetIsAcquisitionData(self):
         return self.__IsAcquisitionData
 
+    def ReadDataSetInfo(self):
+        '''Not implemented for Base class
+        Tries to read info about dataset from given file
+        Will raise specific errors if inputs required for 
+        file type are not set.'''
+        pass
```
paskino, line 761: `raise NotImplementedError` instead of `pass`

Wrappers/Python/examples/error_observer.py
```diff
@@ -0,0 +1,25 @@
...
+An example which shows attaching an error observer and end observer
+to reading a HDF5 file
+'''        
+
+
+if __name__ == "__main__":
+    hdf5_name = r'C:\Users\lhe97136\Work\Data\24737_fd_normalised.nxs'
+    readerhdf5 = cilHDF5ResampleReader()
+    error_obs = ErrorObserver()
+    end_obs = EndObserver(error_observer=error_obs, callback_fn = lambda: iviewer(readerhdf5.GetOutputDataObject(0), readerhdf5.GetOutputDataObject(0)))
```
paskino, line 17: Maybe a comment would help understanding these 2 lines.

paskino (changes requested):
almost ok, just add a comment and a `raise NotImplementedError`

Wrappers/Python/examples/error_observer.py
lauramurgatroyd: Added comments

Wrappers/Python/ccpi/viewer/utils/conversion.py
lauramurgatroyd: The ReadDataSetInfo gets called in the base class because currently the base class is used for reading raw data
So instead of not implemented error, which would break the reading of raw data, I added a check to make sure the user set the size of the dataset and the typecode.
I think we should move the specifics of raw image reading out of the base class to a RawResampleReader, so that the Baseclass is a true baseclass for all readers that is never called itself. Shall I make a separate issue for this?

paskino: Yes please

Wrappers/Python/ccpi/viewer/utils/error_handling.py
```diff
@@ -0,0 +1,51 @@
...
+    def __init__(self, error_observer, callback_fn):
+        '''
+        Parameters
+        ----------
+        error_observer: ErrorObserver
+            the error observer attached to the object the EndObserver is attached to
+        callback_fn: function
+            function to run if an error hasn't been identified by the error_observer'''
+        self.callback_fn = callback_fn
+        self.error_observer = error_observer
```
paskino, line 46: why do you need an error observer in the `EndObserver`? Does it get triggered even if there has been an error?

lauramurgatroyd: Yes. We also tried Exit, Update, UpdateData, SetOutput, none of them worked, they didn't even get triggered when there was no error

PR #189, opened 2022-01-18
Hdf5 cropped reader

### LINKED ISSUE #186: Create a cilHDF5CroppedReader which derives from the cilBaseCroppedReader
Currently we can crop a hdf5 image with code such as:
```python
full_reader = HDF5Reader()
full_reader.SetFileName(self.file_name)
full_reader.SetDatasetName(self.original_image_attrs['dataset_name'])
reader = HDF5SubsetReader()
reader.SetInputConnection(full_reader.GetOutputPort())
reader.SetUpdateExtent(extent)
reader.Update()
image = reader.GetOutputDataObject(0)

```
Should use this code in a cilHDF5CroppedReader so that we can use methods including:
reader.GetElementSpacing()
reader.GetOrigin() etc. as we can with the other readers, so that we have the same interface for all readers

lauramurgatroyd:
Closes #186 

Later we should think about sharing methods between the cropped and resample reader for a filetype (as there is lots of repeated code: #190). Or maybe we combine the cropped and resample reader into a single reader for each filetype?

### DIFF
--- a/Wrappers/Python/ccpi/viewer/CILViewer2D.py
+++ b/Wrappers/Python/ccpi/viewer/CILViewer2D.py
@@ -978,7 +978,7 @@ class CILInteractorStyle(vtk.vtkInteractorStyle):
                 self.SetInitialWindow( window )
 
                 self._viewer.imageSlice.GetProperty().SetColorLevel(self.GetInitialLevel())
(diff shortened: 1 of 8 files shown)


lauramurgatroyd:
Closing this as I have restructured it on the raw_image_readers branch and I'll be opening a new PR there tomorrow

PR #202, opened 2022-02-09
Fix s viewer3d

### LINKED ISSUE #201: viewer3D keypress 's' does not work
`s` should disable the visualisation of the slice in the viewer3D however, since we changed the slice from a `vtkActor` to a `vtkImageSlice` there are still some dangling references to the `vtkActor`  `sliceActor`:

https://github.com/vais-ral/CILViewer/blob/576eeb5de2a6c73871376e79e52baf3029920338/Wrappers/Python/ccpi/viewer/CILViewer.py#L58-L59

https://github.com/vais-ral/CILViewer/blob/576eeb5de2a6c73871376e79e52baf3029920338/Wrappers/Python/ccpi/viewer/CILViewer.py#L235-L238

https://github.com/vais-ral/CILViewer/blob/576eeb5de2a6c73871376e79e52baf3029920338/Wrappers/Python/ccpi/viewer/CILViewer.py#L706-L707

https://github.com/vais-ral/CILViewer/blob/576eeb5de2a6c73871376e79e52baf3029920338/Wrappers/Python/ccpi/viewer/CILViewer.py#L711

paskino:
closes #201


paskino:
I will remove the `colormaps.py` from the PR.

Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -703,12 +703,15 @@ def adjustCamera(self, resetcamera= False):
...
-        self.sliceActor.SetInterpolate(True)
+        # toggle interpolation of slice actor 
+        self._viewer.imageSlice.GetProperty().SetInterpolationTypeToLinear()
         self.renWin.Render()
 
     # Set interpolation off
     def setInterpolateOff(self):
-        self.sliceActor.SetInterpolate(False)
+        # toggle interpolation of slice actor 
+        self._viewer.imageSlice.GetProperty().SetInterpolationTypeToLinear()
```
lauramurgatroyd, line 713: This should be
` self._viewer.imageSlice.GetProperty().SetInterpolationTypeToNearest()`

lauramurgatroyd: (to turn off the interpolation)

paskino: Thanks, I fixed it.

lauramurgatroyd (commented):
Can we change it so we force the objects to become wireframe when we press 's'? so that they don't become opaque when the slice visibility is changed?

paskino:
I tried that, but the result is that something that we may want solid becomes a wireframe. I guess we need to use a different key.

image

paskino:
I guess that for the time being we keep the current state of affairs, i.e. we merge this. Then we change/simplify the interaction and add the Qt bar with buttons to access more complex functionality. What do you think?

lauramurgatroyd:
Okay, makes sense

PR #206, opened 2022-02-22
import setup from setuptools and use pip for installation

lauramurgatroyd:
- use pip to install with conda
- uses setuptools in setup.py script
- formats version string appropriately for pip

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+
+## v22.x.x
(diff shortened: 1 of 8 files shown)


lauramurgatroyd:
Addresses https://github.com/TomographicImaging/CIL/issues/1196 for CILViewer

lauramurgatroyd:
I tested building on windows and it worked fine

paskino (approved):

Issue #165, opened 2021-11-10
create entry_point for executables to resample an image from command line


paskino:
The iDVC app to work from remote requires to resample data on the remote machine. This should be done by command line with the execution of a script or an entry_point.

```yaml

build:
  entry_points:
    - resample = ccpi.viewer.cli.resample:main
    - crop = ccpi.viewer.cli.crop:main
 ```

The executable should accept:
* input filename
* output filename (optional as it could be generated from the input filename)
* target size in kB
* output format

```
resample --input input_file.npy --output output_file.mha --size 1000
crop --input input_file.npy --output output_file.mha --extent 100:200:320:421:402:506
```

**Comment by lauramurgatroyd** [comment]
There may be cases where we want to crop and resample at the same time. We don't have the methods to do this at the moment in a single reader (except in hdf5) but it's something we should do later

**Comment by lauramurgatroyd** [comment]
In the case of a raw file we also need to be given:

- dimensions
- is it fortran
- is it big endian
- type code

In case of a hdf5 file:

- dataset name

For all file types I think they should also set whether its acquisiton data - in this case we don't resample in z direction

**Comment by paskino** [comment]
I guess we could use `docopt` to parse input parameters. Or better `yaml` as it comes with python, so we don't add any dependency?

> For all file types I think they should also set whether its acquisiton data - in this case we don't resample in z direction

This may be a good idea, but I'd code it generically allowing the user to specify `resample_z: true/false`

So in the end the yaml file could be

```yaml
array:
  shape: (1024,1024,1024)
  is_fortran: false
  is_bigendian: true
  typecode: 'float32'
  group_name: /tomosomething # valid for HDF5 and Zarr
resample:
  target_size: 1e7
  resample_z: true
output:
  filename: "this_fname.nxs"
  format: hdf5 # npy, METAImage, NIFTI (or Zarr to come)
```

lauramurgatroyd:
Related to the `resample_z` point:
In our `BaseReader`:
https://github.com/vais-ral/CILViewer/blob/4e41c65a2a86767985f7a352c84090d29af3eacd/Wrappers/Python/ccpi/viewer/utils/conversion.py#L551
there is an option to set whether or not we have acquistition data:
https://github.com/vais-ral/CILViewer/blob/4e41c65a2a86767985f7a352c84090d29af3eacd/Wrappers/Python/ccpi/viewer/utils/conversion.py#L726
[...]

PR #213, opened 2022-03-24
Prevent render pngs from being overwritten

lauramurgatroyd:
Adds incrementing numbers to filename when saving render

Please note I accidentally made these changes on top of those in #212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Adds widget for slicing a volume with the 'c' key in the 3D viewer
 * Add light to 3D viewer
 * Add a few utility methods for volume rendering in 3D viewer
-* Moves saving of render to PNG into utils/io.py
+* Create utils/io.py with code to save current render as PNG
 * Increments number in name of file when saving render to PNG, so multiple captures can be made without them overwriting each other.
 
 ## v22.1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 * Adds widget for slicing a volume with the 'c' key in the 3D viewer
 * Add light to 3D viewer
 * Add a few utility methods for volume rendering in 3D viewer
+* Moves saving of render to PNG into utils/io.py
+* Increments number in name of file when saving render to PNG, so multiple captures can be made without them overwriting each other.
 
 ## v22.1.0
 * requires vtk version >= 9.0.3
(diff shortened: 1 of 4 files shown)


Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -822,12 +831,20 @@ def saveRender(self, filename, renWin=None):
...
+
+        basename = os.path.splitext(filename)[0]
+        first_attempt = True
+        
+        while first_attempt or os.path.exists(saveFilename):
+            str_number = str(self.saved_renders)
+            while len(str_number) < 3:
+                str_number = '0' + str_number
+ 
+            saveFilename = basename + str_number + '.png'
```
paskino, line 845: I think this can be done simply with

```python 
saveFilename = '{}_{:04d}.png'.format(basename, self.saved_renders)
```

Rather than keeping the counter, I was thinking to look in the directory and see if there were files matching the `fname_string` along these lines

```python 
import re, glob

fname_string = '{}_([0-9]*).png'.format(basename)
res_dir = os.path.dirname(filename)

for el in glob.glob(str(res_dir)):
    slist.append(int(re.search(fname_string, str(os.path.basename(el))).group(1)))
```
then `slist` contains a list of the numbers in the `fname_string`: i.e. if the string is `myfile` and in the directory there are `myfile_0000.png` and `myfile_0001.png` it should return 0 and 1. So you now know that the next number needs to be 2.

paskino (changes requested):

paskino (changes requested):
Given that the code to create the filename is entirely the same in both viewers, it makes sense to add it to the `utilis` or, maybe better, create an `io` package that will be used also for #165

lauramurgatroyd:
Updated to move `SaveRenderToPNG` into `utils/io.py`

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 * Adds widget for slicing a volume with the 'c' key in the 3D viewer
 * Add light to 3D viewer
 * Add a few utility methods for volume rendering in 3D viewer
+* Moves saving of render to PNG into utils/io.py
```
paskino, line 9: I'd say create `io` module

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -0,0 +1,45 @@
...
+        render window to save contents of.
+    filename: str
+        name of file to write PNG to.
+    '''
+    w2if = vtk.vtkWindowToImageFilter()
+    w2if.SetInput(render_window)
+    w2if.Update()
+
+    basename = os.path.splitext(os.path.basename(filename))[0]
+    # Note regex matching is different to glob matching:
```
paskino, line 25: would that be a problem?

paskino (approved):

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: it's not a problem it just explains why there are both the `regex` and the `fname_string` variables which look very similar

PR #216, opened 2022-04-07
Command line tool for resampling a dataset and writing it out to a new file

lauramurgatroyd:
[resample.py](https://github.com/vais-ral/CILViewer/blob/resampling_cli/Wrappers/Python/ccpi/viewer/cli/resample.py) requires a file name and yaml file, and reads a dataset from the file, writing out a downsampled version to a HDF5 file or a metaimage file. Alternatively, it can just take a filename and some command line args.
The comments in resample.py show:
An example input yaml file:
https://github.com/vais-ral/CILViewer/blob/a45015eb0e6c5d6bc1fc6f29f4763fbd0a86cd95/Wrappers/Python/ccpi/viewer/cli/resample.py#L11-L25

And an example of the structure of the hdf5 file written out:
https://github.com/vais-ral/CILViewer/blob/14a3ba88ef2031368f035e832812029a48c7e28d/Wrappers/Python/ccpi/viewer/cli/resample.py#L39-L67

To do this it uses the following new classes:

https://github.com/vais-ral/CILViewer/blob/a45015eb0e6c5d6bc1fc6f29f4763fbd0a86cd95/Wrappers/Python/ccpi/viewer/utils/io.py#L80-L89

https://github.com/vais-ral/CILViewer/blob/a45015eb0e6c5d6bc1fc6f29f4763fbd0a86cd95/Wrappers/Python/ccpi/viewer/utils/io.py#L388-L421

https://github.com/vais-ral/CILViewer/blob/a45015eb0e6c5d6bc1fc6f29f4763fbd0a86cd95/Wrappers/Python/ccpi/viewer/utils/io.py#L540-L544

https://github.com/vais-ral/CILViewer/blob/a45015eb0e6c5d6bc1fc6f29f4763fbd0a86cd95/Wrappers/Python/ccpi/viewer/utils/io.py#L630-L638

Also see: https://github.com/vais-ral/CILViewer/blob/resampling_cli/Wrappers/Python/examples/image_reader_and_writer.py
which provides examples of using the above classes


lauramurgatroyd:
Possible changes which may be made in this pull request:
- Restructuring the ImageReader to be more VTK-like

Future changes (not in this PR):
-  Supporting resampling tiffs
-  A similar script for cropping - the functionality is already there for cropping in the ImageReader, just didn't want to make this PR any longer!

lauramurgatroyd:
@paskino suggested restructuring the ImageReader to be more VTK-like.
Currently it is called with something like:
https://github.com/vais-ral/CILViewer/blob/855265cf3b13339ee2e965e99c17c17a49018b54/Wrappers/Python/examples/image_reader_and_writer.py#L64-L69

We could instead have:

```
reader = ImageReader()
reader.SetFileName(DATASET_TO_READ)
reader.SetTargetSize(TARGET_SIZE)
reader.SetResampleZ(True)
reader.SetLogFile(LOG_FILE)

resampled_image = reader.read()
resampled_image_attrs = reader.get_loaded_image_attrs()
original_image_attrs = reader.get_original_image_attrs()
```

However, I don't think we should make ImageReader derive from VTKPythonAlgorithmBase.

This is because basically all the ImageReader does is identify which reader to use, attaches observers to it, to deal with error handling and progress reporting, and when reading, just calls Update() on the specific reader and updates the attributes appropriately.

Currently, you could call ImageReader.read() in a thread, passing in progress_callback and it would do the updating of progress for you.

If we made ImageReader derive from VTKPythonAlgorithmBase then the user would have to set up an observer on ImageReader to handle progress reporting, and inside ImageReader this would be attaching the observer to the specific readers - so it seems like we unnecessarily have 2 observers instead of one and it's more work for the user.

But if we don't have it derive from VTKPythonAlgorithmBase then would the sugested new VTK-like structure of the ImageReader make it confusing (in that you call read() instead of Update())?

Let me know your thoughts / whether the above makes any sense

paskino:
I don't believe it should be inheriting from `vtkPythonAlgorithmBase` I was suggesting that it could behave/look like one.

Some points of discussion:

* I am not a fan of `set_up` as we have in CIL, and I prefer more methods with self explanatory names, but it's matter of taste  
* The viewer is a bit inconsistent regarding the use of camel case: sometimes we use the VTK convention (e.g. `SetFileName`), sometimes the Qt convention (e.g. `setFileName`). Here you used the CIL convention (e.g. `set_file_name`). Shall we choose one?
* I would go for the VTK convention, because this class returns a VTK object. Clearly the viewer has both a Qt soul and a VTK soul. Which convention do we choose overall?
* It's still good to retain some flexibility and be able to initialise our classes as discussed in [CIL](https://tomographicimaging.github.io/CIL/nightly/developer_guide.html#conventions-on-new-cil-objects) with the obvious change in method naming and presence here of getters rather than properties.

paskino:
CIL docs say
> the creator of a class should require the essential and often-configured parameters as named parameters.

So you could add the dataset name to the creator.

Wrappers/Python/ccpi/viewer/utils/hdf5_io.py
```diff
@@ -55,64 +58,83 @@ def __init__(self):
...
+                    if not (lname in f):
+                        raise Exception("No dataset named {} exists in {}.".format(lname, self._FileName))
+            self._DatasetName = lname
 
     def GetDatasetName(self):
-        return self.__DatasetName
+        return self._DatasetName
 
     def Set4DIndex(self, index):
         '''Sets which index to read, in the case of a 4D dataset'''
```
paskino, line 119: Does this mean that this will only return a 3D object?
I find this index a bit confusing. Is it used to tell which index not to read?

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -43,3 +71,691 @@ def SaveRenderToPNG(render_window, filename):
...
+            reader = cilHDF5ResampleReader()
+
+        reader.SetDatasetName(self._HDF5DatasetName)
+
+        return reader
+
+    def _SetUpLogger(self, fname):
+        """Set up the logger """
+        self.logger = logging.getLogger("ccpi.viewer.utils.io.ImageReader")
+        self.logger.setLevel(logging.INFO)
```
paskino, line 402: I guess the logging level should be settable?

Wrappers/Python/ccpi/viewer/utils/hdf5_io.py
lauramurgatroyd: yes, only a 3D object is returned, so it's selecting which index in the 4th dimension to read

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -43,3 +71,691 @@ def SaveRenderToPNG(render_window, filename):
...
+                    # with shape flipped and pretend the array is
+                    # C order.
+                    array = array.reshape(data.GetDimensions()[::-1])
+                else:
+                    array = None
+                try:
+                    if array is None:
+                        dset = f.create_dataset(dataset_name, dataset_info['shape'])
+                    else:
+                        dset = f.create_dataset(dataset_name, data=array)
```
paskino, line 676: As an enhancement, I think it makes sense to use [chunking ](https://github.com/paskino/CIL/blob/3105be92c69fbc54c976e729030f85a1144c3fb6/Wrappers/Python/cil/io/NEXUSDataWriter.py#L139-L155)here.

An example of usage of that functionality can be found [here](https://github.com/paskino/CIL-work/blob/db4d4d227eacc86c9111992359c147492594db85/PSDI/test_chunking.py#L66-L74)

Wrappers/Python/conda-recipe/meta.yaml
```diff
@@ -32,6 +32,7 @@ requirements:
     - eqt
     - importlib_metadata    # [py<38]
     - h5py
+    - schema
```
paskino, line 35: Do we really need `schema`? Can we just do without validation? The user is responsible for passing the right stuff?

paskino (commented):
Discussion whether `schema` is needed. Also there is no unit test for the executable.

Wrappers/Python/conda-recipe/meta.yaml
lauramurgatroyd: I think it makes it more user-friendly

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: Ok I will have a go!

paskino: yes. It doesn't change speed if we read the whole dataset, but it could massively if we read only one slice at a time.

Wrappers/Python/ccpi/viewer/utils/hdf5_io.py
paskino: To make it ~more complicated~ clearer, I'd use the complementary information, i.e. request what information we actually need/want.

lauramurgatroyd: Sorry what do you mean by this?

lauramurgatroyd:
I have added in chunking when writing to HDF5 and made a unit test for the resample entry point.
I am not sure about the names for the new classes I have made, especially some include 'vortex' - are we set on this yet?

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -43,3 +71,761 @@ def SaveRenderToPNG(render_window, filename):
...
+        for i in range(0, len(self._ChildDatasets)):
+            writer.AddChildDataset(self._ChildDatasets[i],  self._ChildDatasetsAttributes[i])
+        return writer
+
+    def _GetMetaImageWriter(self):
+        writer = vtk.vtkMetaImageWriter()
+        writer.SetInputData(self._ChildDatasets[0])
+        return writer
+
+class vortexHDF5ImageWriter(ImageWriter):
```
lauramurgatroyd, line 644: should we name this something else? Are we committed to renaming the module to vortex?

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -43,3 +71,761 @@ def SaveRenderToPNG(render_window, filename):
...
+                                self._FileName, dataset_name))
+
+                if entry_num != 1:
+                    dset.attrs['original_dataset'] = 'entry1/tomo_entry/data/data'
+
+                for key, value in dataset_info.items():
+                    dset.attrs[key] = value
+
+
+class vortexHDF5ImageReader(HDF5Reader):
```
lauramurgatroyd, line 759: should we name this something else? Are we committed to renaming the module to vortex?

lauramurgatroyd:
I think I made the requested changes, not sure about the names of the reader/writer for our new format of nexus/hdf5 file

Wrappers/Python/ccpi/viewer/utils/io.py
paskino: I don't know if we need `vortex` in the class name. I think it could be removed. It will be found in the `vortex` package anyway.

Is the path to the data `entry{}/tomo_entry/data/data` not part of the NeXuS specification?
The class name is now `HDF5ImageReader`, what is it meant by `Image`?

lauramurgatroyd: well the thing is we already have a HDF5 reader which is called HDF5Reader, so it shouldn't have the same name as that. This class is for a HDF5 type that we have defined that we expect to include certain info - the origin and spacing.
Kind of the viewer's equivalent of the CIL Nexus reader expecting certain attributes etc.

lauramurgatroyd: yes you are right, `entry{}/tomo_entry/data/data` is part of the NeXuS specification.
We do not have a specific NeXuS reader in the viewer yet just a generic HDF5 reader.
We should have a NeXuS one, and then this could inherit from the NeXuS one. I can either do that now or in another PR.

Wrappers/Python/ccpi/viewer/cli/resample.py
```diff
@@ -0,0 +1,135 @@
...
+  is_fortran: False
+  is_big_endian: True
+  typecode: 'float32'
+  dataset_name: '/entry1/tomo_entry/data/data' # valid for HDF5 and Zarr
+resample:
+  target_size: 100000
+  resample_z: True
+output:
+  file_name: 'this_fname.nxs'
+  format: 'hdf5' # npy, METAImage, NIFTI (or Zarr to come)
```
paskino, line 36: I think the yaml file is good, however I'd like to have a command line tool which can be called from the command line:

```bash 
resample -i input_file.h5 -o resampled.h5 --target_size 100M --resample_z 1
```

This should parse the parameters, read the metadata in the input file, generate the meta.yaml and execute. 

Raw files are the only ones that require lots of parameters and it may be awkward to have a very long command line to specify all of these. In that case we could have a command like `resample --example` which generates an example basic yaml file

paskino: or we could support 
```bash

resample -i config.yaml
```

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: What do we rename to now that we keep cilviewer as the name? Maybe cilviewerHDF5Reader?

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: rename to cilviewer or clv

lauramurgatroyd:
The docker tests fail because the io.py file tries to import the cilviewer version, which can't be found.

When the docker tests were added, the test for the version was skipped as we couldn't understand why it was failing.

We need to fix this.

https://github.com/vais-ral/CILViewer/issues/282

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: renamed to cilviewer

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: renamed to cilviewerHDF5Reader

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+                raise Exception("file_name, dataset(/s) and attribute(/s), are required.")
+        for var in [self._ChildDatasets, self._ChildDatasetsAttributes]:
+            if var is []:
+                raise Exception("child dataset(/s) and attribute(/s), are required.")
+
+        with h5py.File(self._FileName, 'w') as f:
+
+            # give the file some important attributes
+            f.attrs['file_name'] = self._FileName
+            f.attrs['viewer_version'] = version
```
lauramurgatroyd, line 686: The reason the docker github action is failing is because in io we are importing version, and saving it to the file we create here. for now we could not do this and it would stop the action from failing? I don't know how to fix the action.

Wrappers/Python/ccpi/viewer/cli/resample.py
```diff
@@ -0,0 +1,236 @@
...
+Supported file types for reading:
+hdf5, nxs, mha, raw, numpy
+
+Supported file types for writing:
+hdf5, nxs, mha
+
+'''
+
+EXAMPLE_YAML_FILE = '''
+EXAMPLE INPUT YAML FILE:
```
paskino, line 22: Is this a valid yaml file with this line?

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+
+    def Read(self, *args, **kwargs):
+        ''' reads self._FileName
+            returns vtkImageData'''
+        # identifies file type
+        # uses appropriate reader based on file type and cropping or resampling
+
+        if self._Crop:
+            if self._TargetZExtent is None:
+                raise Exception("Error: if crop is set to True, target_z_extent must be set.")
```
paskino, line 235: Is it a `TypeError`?

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+            if self._Resample:
+                warnings.warn(
```
paskino, line 237: this should be done in with `self.logger.warning`

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+        raw_attrs_schema = Schema({
+            'shape': Or(list, np.ndarray, tuple),
+            'is_fortran': bool,
+            'is_big_endian': bool,
+            'typecode': str
+        })
+        raw_attrs_schema.validate(raw_attrs)
+        return raw_attrs
+
+    def _GetReader(self, progress_callback=None):
```
paskino, line 277:
```
'''Returns an appropriate reader for the image file provided

The appropriate reader is decided by matching the extension of the file to be read, which can be 
mha, mhd for METAIO files, npy for numpy file format, raw for binary blobs, nxs h5 or hdf5 for HDF5 files,
tiff or tif for TIFF stacks.
No actual check of the file format is performed.
'''
```

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+                 raw_image_attrs=None,
+                 hdf5_dataset_name="entry1/tomo_entry/data/data",
+                 log_file=None):
+        '''
+        Constructor
+
+        Parameters
+        ----------
+        file_name: os.path or string, default None
+            file name to read
```
paskino, line 103: In case of TIFF files, either the directory to the TIFF files, or the path to one such file is needed.

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+# TODO:
+# supporting tiffs
+# write out other filetypes
+
+
+class ImageReader(object):
+    '''
+    Generic reader for reading to vtkImageData
+    Currently reads: HDF5, MetaImage, Numpy, Raw
+    Later will support TIFFs
```
paskino, line 80: It seems to support TIFFs!!

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+            if len(image_files) == 0:
+                raise Exception('No tiff files were found in: {}'.format(self._FileName))
+            self._data = self._GetTiffImageReader(image_files)
+
+        if file_extension not in ['.tif', '.tiff']:
+            # currently the tiff reader doesn't take these inputs:
+            reader.SetFileName(self._FileName)
+            # if we have acquisition data then we would not resample on
+            # the z axis:
+            reader.SetIsAcquisitionData(not self._ResampleZ)
```
paskino, line 314: This is an interesting learned guess.

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+        reader.AddObserver(vtk.vtkCommand.ErrorEvent, error_obs)
+        # Could add end observer so that we don't continue to do anything
+        # else if an error does occur in reader?
+
+        return reader
+
+    def _GetMetaImageReader(self, progress_callback=None):
+        if self._Crop:
+            reader = cilMetaImageCroppedReader()
+            reader.SetTargetZExtent(tuple(self._TargetZExtent))
```
paskino, line 341: all readers, with the https://github.com/vais-ral/CILViewer/blob/4cc7472bb0e11a79f6ae163894f66068568b6a59/Wrappers/Python/ccpi/viewer/utils/conversion.py#L1900

support `SetTargetZExtent` which could be moved in the `Read` method, at about line 315

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+        else:
+            reader = cilMetaImageResampleReader()
+        return reader
+
+    def _GetNumpyImageReader(self, progress_callback=None):
+        if self._Crop:
+            reader = cilNumpyCroppedReader()
+            reader.SetTargetZExtent(tuple(self._TargetZExtent))
```
paskino, line 349: As above

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+        self._OriginalImageAttrs['file_name'] = self._FileName
+        self._OriginalImageAttrs['resampled'] = False
+        self._OriginalImageAttrs['cropped'] = False
+
+
+class ImageWriter(object):
+    '''
+    Writer for writing out a modified i.e. resampled or cropped dataset.
+    Currently supports writing to HDF5 and metaimage.
+    Will later support other formats e.g zarr
```
paskino, line 456: Do we need to state this?

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+            raise Exception("File format is not supported. Supported types include hdf5/nexus.")
+
+        writer.SetFileName(self._FileName)
+        return writer
+
+    def _GetHDF5Writer(self):
+        writer = cilviewerHDF5Writer()
+        writer.SetOriginalDataset(None, self._OriginalDatasetAttributes)
+        writer.SetChunking(self._Chunking)
+        writer.SetChunkShape(self._ChunkShape)
```
paskino, line 638: What is the chunk shape set to? It seems we ask for chunking but not set the shape.

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+            writer.AddChildDataset(self._ChildDatasets[i], self._ChildDatasetsAttributes[i])
+        return writer
+
+    def _GetMetaImageWriter(self):
+        writer = vtk.vtkMetaImageWriter()
+        writer.SetInputData(self._ChildDatasets[0])
+        return writer
+
+
+class cilviewerHDF5Writer(ImageWriter):
```
paskino, line 650: Why is this in this file rather than in https://github.com/vais-ral/CILViewer/blob/master/Wrappers/Python/ccpi/viewer/utils/hdf5_io.py

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +66,765 @@ def SaveRenderToPNG(render_window, filename):
...
+                          "Dataset with name {1} already exists in this file.".format(self._FileName, dataset_name))
+
+                if entry_num != 1:
+                    dset.attrs['original_dataset'] = 'entry1/tomo_entry/data/data'
+
+                for key, value in dataset_info.items():
+                    dset.attrs[key] = value
+
+
+class cilviewerHDF5Reader(HDF5Reader):
```
paskino, line 761: same as above.

paskino (changes requested):
All in all is OK, but there are a few points here and there that needs addressing.

Use `cilTIFFCroppedReader` and `cilTIFFResampleReader` where possible.

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: updated

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: There is a method SetChunkShape

lauramurgatroyd: (in the ImageWriter) I mean. By default this is None. If the chunk shape is None in the cilviewerHDF5Writer then it uses one slice on the z axis as the chunk shape by default

lauramurgatroyd: (if chunking has been set to True)

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +58,772 @@ def SaveRenderToPNG(render_window, filename):
...
+        self._Chunking = chunking
+
+    def SetChunkShape(self, chunk_shape):
+        '''
+        Parameters
+        ----------
+        chunk_shape: tuple, default: one slice on the z axis
+            If SetChunking() has been set to True, this will be
+            the size of a chunk written to the file.
+        '''
```
lauramurgatroyd, line 590: @paskino this is where the chunk shape is set and default explained

Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -42,3 +58,772 @@ def SaveRenderToPNG(render_window, filename):
...
+        reader.SetIsFortran(isFortran)
+        reader.SetTypeCodeName(typecode)
+        reader.SetStoredArrayShape(shape)
+
+        return reader
+
+    def _GetHDF5ImageReader(self):
+        if self._Crop:
+            reader = cilHDF5CroppedReader()
+            reader.SetTargetExtent([0, -1, 0, -1, self._TargetZExtent[0], self._TargetZExtent[1]])
```
lauramurgatroyd, line 394: The HDF5 reader has differing behaviour with regards to setting the target extent - SetTargetExtent is used instead of SetTargetZExtent

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: I didn't do it for this reason: https://github.com/vais-ral/CILViewer/pull/216/files/0944550751358d7549db2a3b8ebadc4877d6c303#r944597190

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: Again, see https://github.com/vais-ral/CILViewer/pull/216/files/0944550751358d7549db2a3b8ebadc4877d6c303#r944597190

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: no

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: @paskino, shall I just not print the version here for now so that the action passes?

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: For circular import reasons it needs to stay in the io file

lauramurgatroyd: That is unless we move what it inherits from elsewhere

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: I have also made a new class called ImageWriterInterface and moved methods used by both the ImageWriter and cilviewerHDF5Writer into there to make things tidier, more logical, and less confusing.

lauramurgatroyd:
I think I have responded to all the comments by either making the requested changes or replying to the comments.

The Tiff tests have broken again and I believe it's something to do with VTK8 - running locally with VTK9 everything passes. So I have that to fix with VTK8 on Monday, but otherwise it is ready to re-review.

lauramurgatroyd:
The problem was not with VTK8! It was with glob sorting the files it found differently on windows and linux!! So in the action it was reading the tiff files in the wrong order. So I added some sort methods, the ones used in CIL, to resolve this issue.

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: I updated the comment to explain better about what is happening. Also made this issue https://github.com/vais-ral/CILViewer/issues/309

paskino (approved):
Apart from a docstring, I think it's all in good shape.

There is a [suggestion ](https://github.com/vais-ral/CILViewer/pull/216#discussion_r947590169)on how I'd change the docstring to explain a little better the chunking and the default values

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: That's right

Wrappers/Python/ccpi/viewer/utils/io.py
lauramurgatroyd: That's what I've done for now, will make a linked issue later

Wrappers/Python/test/test_io.py
```diff
@@ -0,0 +1,301 @@
...
+
+    target_image_shape = (int(xy_axes_magnification * shape[0]), int(xy_axes_magnification * shape[1]), num_chunks)
+    return target_image_shape
+
+
+# TODO:
+# First, focus on resampling and writing out to HDF5
+# for this we need:
+# test image attributes given by reader
+# test writing out a file
```
lauramurgatroyd, line 28:
```suggestion

```

PR #221, opened 2022-04-12
Fix slicing of downsampled data by setting extent in world coordinates

### LINKED ISSUE #219: Slice widget does not work correctly on downsampled image data
This is because it uses the image coordinates for the sizing and positioning of the widget, instead of the world coordinates

### LINKED ISSUE #220: 3D viewer's interactor style has no methods to convert between image and world coordinates
In the 2D viewer's interactor style we have methods such as:
https://github.com/vais-ral/CILViewer/blob/b218cb63fbf710c0bc5c5880578560eba6fe372e/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L858-L877

We need to use these in the 3D viewer as well.

When we address #214 we should have these methods in the base class.

Until then, we should copy the methods needed to fix #219 into the 3D viewer.

lauramurgatroyd:
Closes #219 and closes #220

### DIFF
--- a/Wrappers/Python/ccpi/viewer/CILViewer.py
+++ b/Wrappers/Python/ccpi/viewer/CILViewer.py
@@ -264,7 +264,8 @@ class CILInteractorStyle(vtk.vtkInteractorStyleTrackballCamera):
                 planew = vtk.vtkImplicitPlaneWidget2()
                 
                 rep = vtk.vtkImplicitPlaneRepresentation()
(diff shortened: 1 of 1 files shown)


paskino (approved):
I suppose we only tried this with spacing of 1 and origin in 0,0,0 and problems arose once downsampled data was shown.

I had not understood that the extent of in world coordinates, but it makes sense.

PR #222, opened 2022-04-12
Improvements to iviewer example

lauramurgatroyd:
Allows selection of viewer types in the case that 2 images are displayed. This means we can view an image in both 2D and 3D at the same time - in the example when iviewer.py is run, it does this with the head image.

Changes the default VolumeOpacityMethod in the 3D viewer created in the example (if present) to 'scalar' - this is optimal for the head dataset - as with the default 'gradient' method, the volume render can't be seen - only the teeth very faintly!


paskino (approved):
I think this is fine. The opacity setting will need to be selectable, and modifiable at runtime.

 I suppose this would be something for the `QFrame` where the viewer is embedded, with some button and popup as we have discussed other times.

Issue #218, opened 2022-04-11
Can we ship an example image with the CILViewer?


lauramurgatroyd:
Almost all of the [examples](https://github.com/vais-ral/CILViewer/tree/master/Wrappers/Python/examples) require an image file.

Like with CIL, can we ship some example image with CILViewer so the examples can easily be run?

paskino:
Sure. The head image is from [VTK Data](https://github.com/open-cv/VTKData/tree/master/Data/headsq). Check the license and ship it!

lauramurgatroyd:
The only license I could find was this, which seems to just refer to the software, not specifically data

https://github.com/open-cv/VTKData/blob/master/Copyright.txt

lauramurgatroyd:
closed by #223

PR #223, opened 2022-04-19
Add example of running the 2D and 3D viewers without Qt

lauramurgatroyd:
- Adds examples of running the 2D and 3D viewers without Qt
- Uses data from the CIL-Data repo by adding this as a run requirement
- Adds classes to easily read the head dataset from the CIL-Data repo
- A unit test for reading the head dataset from the CIL-Data repo


paskino (changes requested):
I think this is a good idea, however we need to provide the user with the data to load. 

The head is from [VTKData](https://github.com/naucoin/VTKData/tree/master/Data/headsq), but it's not read with the MetaImage reader but with something like

```python
reader = vtk.vtkImageReader2()
reader.SetFilePrefix(path_to_files + "/Data/headsq/quarter")
reader.SetDataExtent(0, 63, 0, 63, 1, 93)
reader.SetDataSpacing(3.2, 3.2, 1.5)
reader.SetDataOrigin(0.0, 0.0, 0.0)
reader.SetDataScalarTypeToUnsignedShort()
reader.UpdateWholeExtent()
```

lauramurgatroyd:
I have added:

- The head dataset from VTKData
- The license file relating to it: https://github.com/vais-ral/CILViewer/blob/8d57816d933906d5c80c96703e1cfcdba644511c/Wrappers/Python/ccpi/viewer/data/headsq/LICENSE.txt - I am not sure if I should modify this somehow to indicate that it is only referring to the head data and not other files in the repo.
- A way to easily load this dataset: https://github.com/vais-ral/CILViewer/pull/223/files#diff-11423bd61586ed8a9dd4b4ab42d9e6978672812ac66f319fa8f6bc6838a78ec1 - heavily inspired by: https://github.com/TomographicImaging/CIL/blob/master/Wrappers/Python/cil/utilities/dataexample.py
- A test for reading the dataset
- Used the head dataset in the example of running the 2D viewer

Closes: #218

paskino (changes requested):
I think it is a good idea. Why not use (compressed) MHD/MHA? I'd add the VTKData license and a README file in the directory of the data where it is described that the saved file comes from VTKData.

Also, would we not plan to have more example data? In this case, it makes sense to host the data into a separated repository, and configure it as a submodule, as done with [CIL](https://github.com/TomographicImaging/CIL#getting-the-code).

lauramurgatroyd:
I was going to use mha/mhd until your previous comment, but yes we can do that :) yes we would plan to have more data. I will try to copy what is done in CIL.

paskino:
Sorry, I meant to say that we needed to cite the origin of the data. And since the data is not in MHA/MHD format, I commented about how it was possible to pass from data you find in VTKData to the file we always use.

lauramurgatroyd:
I think that would be a good idea if others are happy with it. Would we then have 2 license files in there or combine them somehow indicating which datasets they refert to?

lauramurgatroyd:
Note: this is waiting on https://github.com/TomographicImaging/CIL-Data/pull/8

lauramurgatroyd:
I am not sure we need a submodule, just having CIL-Data as a requirement seems sufficient?

.github/workflows/README.md
```diff
@@ -6,8 +6,3 @@ This github action builds and tests the conda package, by using the [conda-packa
 When making an [annotated](https://git-scm.com/book/en/v2/Git-Basics-Tagging) tag, the code is built, tested and published to the [ccpi conda channel for ccpi-viewer](https://anaconda.org/ccpi/ccpi-viewer/files) as noarch package, as pure python. Previously we provided packages for linux, windows and macOS versions.
 
 When pushing to master, or opening or modifying a pull request to master, a single variant is built and tested, but not published. This variant is `python=3.7` and `numpy=1.18`. Notice that we removed the variants from the recipe as we do not have a strong dependency on numpy or python.
```
lauramurgatroyd, line 8:
```suggestion
When pushing to master, or opening or modifying a pull request to master, a single variant is built and tested, but not published. This variant is `python=3.7` and `numpy=1.18`. Notice that we removed the variants from the recipe as we do not have a strong dependency on numpy or python.

## Auto-yapf: [autoyapf](https://github.com/vais-ral/CILViewer/blob/master/.github/workflows/autoyapf.yml)
This action runs yapf auto-formatting on the whole repository, it runs on both push to master, and push to pull requests. Once it has ran yapf, it will commit to the branch it has checked out for the action, so on master, and the pull request branch when running on a pull request.

It works by using mritunjaysharma394's autoyapf github action, to perform the formatting then manually committing to the branch that is checked out, but only after checking if there are any changes to files in the git tree.
```

Wrappers/Python/conda-recipe/meta.yaml
```diff
@@ -34,6 +34,7 @@ requirements:
     - eqt
     - importlib_metadata    # [py<38]
     - h5py
+    - cil-data >= 22.0.0
```
lauramurgatroyd, line 37:
```suggestion
    - cil-data >=22.0.0
```

Wrappers/Python/ccpi/viewer/utils/example_data.py
```diff
@@ -0,0 +1,63 @@
...
+        return TestData.HEAD
+
+
+class TestData(object):
+    '''Class to return test data
+    
+    provides 1 dataset:
+    HEAD = 'headsq.mha'
+    '''
+    HEAD = 'headsq.mha'
```
lauramurgatroyd, line 36:
```suggestion
    HEAD = 'head.mha'
```

Wrappers/Python/ccpi/viewer/utils/example_data.py
```diff
@@ -0,0 +1,63 @@
...
+    @classmethod
+    def dfile(cls):
+        return TestData.HEAD
+
+
+class TestData(object):
+    '''Class to return test data
+    
+    provides 1 dataset:
+    HEAD = 'headsq.mha'
```
lauramurgatroyd, line 34:
```suggestion
    HEAD = 'head.mha'
```

.gitignore
```diff
@@ -1,6 +1,7 @@
 ccpi_viewer.egg-info
 *.pyc
 __pycache__
+*.log
```
paskino, line 4: is this necessary?

Wrappers/Python/ccpi/viewer/utils/example_data.py
```diff
@@ -0,0 +1,63 @@
+import os
+import os.path
+import vtk
+import sys
+
+# this is the default location after pip install
```
paskino, line 6: I suppose this comment should relate to the location of the data coming from `cil-data`

paskino (approved):
It's fine. Maybe update one comment as per review.

.gitignore
lauramurgatroyd: well, the viewer often makes log files which we don't want to commit

Wrappers/Python/ccpi/viewer/utils/example_data.py
lauramurgatroyd, line 6:
```suggestion
# this is the location where the cil-data module saves the datasets
```

paskino (approved):

Issue #217, opened 2022-04-11
Decide on naming conventions for the viewer.


lauramurgatroyd:
See https://github.com/vais-ral/CILViewer/pull/216#issuecomment-1093772185

paskino:
possibly similar comments are in #215

paskino:
I'm now inclined to think that we should mainly adhere to VTK's conventions, with the exception of adopting Qt's conventions when we are dealing with qt stuff.

Possibly this also would mean that we make submodules like

```
vortex
vortex.qt
vortex.web #trame
vortex.io
vortex.utils
```

lauramurgatroyd:
There are basically 3 different conventions we should/could follow:

For each I use example names to show how they are written (case-wise/underscores etc.)

1. VTK convention:
   Class name: `vtkCellArray()`
   Method names: `.GetValue()`

2. Qt convention:
   Class names: `QComboBox`
   Method names: `.addItem()`

3. CIL / Python convention:
   Class names: `BlockDataContainer`
   Method names: `.get_item()`

lauramurgatroyd:
The question is where do we use each of these?

1. **Readers.** These are VTK-like classes so use VTK style. At the moment the prefix used is cil. Should we change this to cilviewer or cilv or cilvie to avoid confusion?
https://github.com/vais-ral/CILViewer/blob/c9d8604ef7accd5d6eec68b888f2a4396254270c/Wrappers/Python/ccpi/viewer/utils/conversion.py#L567-L568

2. **Viewers**
   https://github.com/vais-ral/CILViewer/blob/c9d8604ef7accd5d6eec68b888f2a4396254270c/Wrappers/Python/ccpi/viewer/CILViewer.py#L425

At the moment these use the Qt convention of camelCase methods names, even though they are completely separate from Qt. So we should consider a change although it will be a big backwards-compatability break. If we change, what do we change to? VTK-Style? Does that mean we would also change the name of the viewers to have a prefix?

3. **Viewer interactor style**
Again, Qt/Python style name, but VTK-style methods - i.e. methods are all UpperCase words

4. **Qt windows with viewers embedded**
Has Qt style as expected:
https://github.com/vais-ral/CILViewer/blob/master/Wrappers/Python/ccpi/viewer/QCILViewerWidget.py

But iviewer and other viewer windows in this file, don't follow the Qt style, whereas perhaps they should?
[...]

PR #224, opened 2022-04-20
Create pull_request_template.md

lauramurgatroyd:
Mostly borrowed from @gfardell


lauramurgatroyd:
Not sure whether to link the CIL developer guide or not as we follow *some* of it but not all - as most classes are VTK-style!

paskino:
It's a good idea. I guess we should decide about the viewer conventions: for instance I think we use Qt's conventions for module names `verbSomeStuff` while VTK uses `VerbSomeStuff`, but this is not quite uniform throughout the code.

paskino (changes requested):
See https://github.com/vais-ral/CILViewer/pull/224#issuecomment-1119367671

lauramurgatroyd:
converting to draft until we have decided on https://github.com/vais-ral/CILViewer/issues/217

lauramurgatroyd:
Some info was added about coding style, think it's ready to review again.

paskino (approved):

Issue #227, opened 2022-04-27
updateVolumePipeline does not update the opacity method


lauramurgatroyd:
Here:
https://github.com/vais-ral/CILViewer/blob/353eee677cf02353e8d045afb3da3921e302d1f7/Wrappers/Python/ccpi/viewer/CILViewer.py#L838

We should set the opacity like we do when we install the volume pipeline:
https://github.com/vais-ral/CILViewer/blob/353eee677cf02353e8d045afb3da3921e302d1f7/Wrappers/Python/ccpi/viewer/CILViewer.py#L690-L693

PR #228, opened 2022-04-27
Fix updating of volume opacity method in 3D viewer

lauramurgatroyd:
- Closes: #227 
- Fixes the way the gradient opacity is set so that it is mapping the values in the image gradient
- Adds methods for setting and getting the window or percentiles for:
   - The colour
   - The scalar opacity (used if setVolumeRenderOpacityMethod is set to scalar)
   - The gradient opacity (used if setVolumeRenderOpacityMethod is set to gradient)
   https://github.com/vais-ral/CILViewer/blob/03701d1949dde7f42b66a1c45f86bef337f108c6/Wrappers/Python/ccpi/viewer/CILViewer.py#L775-L886
- Adds method for getting the full range of image or gradient values: https://github.com/vais-ral/CILViewer/blob/03701d1949dde7f42b66a1c45f86bef337f108c6/Wrappers/Python/ccpi/viewer/CILViewer.py#L968-L977

- Makes an example with dropdown for updating opacity method:
image
image

**Note backwards compatibility break:
setVolumeColorLevelWindow  has been renamed to: setVolumeColorWindow**

### DIFF
--- a/Wrappers/Python/ccpi/viewer/CILViewer.py
+++ b/Wrappers/Python/ccpi/viewer/CILViewer.py
@@ -494,6 +494,9 @@ class CILViewer():
         volumeProperty = vtk.vtkVolumeProperty()
         self.volume_property = volumeProperty
 
+        # These may be optionally set by the user:
+        self.volume_colormap_limits = None
+
         
         # The volume holds the mapper and the property and
         # can be used to position/orient the volume.
@@ -680,20 +683,16 @@ class CILViewer():
         self.renWin.Render()
 
     def installVolumeRenderActorPipeline(self):
-        
         self.volume_mapper.SetInputData(self.img3D)
 
         # define colors and opacity with default values
-        colors, opacity = self.getColorOpacityForVolumeRender()
+        colors, opacity = self.getColorOpacityForVolumeRender(method=self.getVolumeRenderOpacityMethod())
 
         self.volume_property.SetColor(colors)
 
-        self.setDefaultScalarOpacity(self.volume_property.GetScalarOpacity())
-        self.setDefaultGradientOpacity(self.volume_property.GetGradientOpacity())
+        self.setDefaultScalarOpacity()
 
-        if self.getVolumeRenderOpacityMethod() == 'gradient':
-            self.volume_property.SetGradientOpacity(opacity)
-        elif self.getVolumeRenderOpacityMethod() == 'scalar':
+        if self.getVolumeRenderOpacityMethod() == 'scalar':
             self.volume_property.SetScalarOpacity(opacity)
         else:
             # currently this is not relevant, but in the future one may want to do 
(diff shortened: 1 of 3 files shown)


Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -687,6 +687,10 @@ def installVolumeRenderActorPipeline(self):
         colors, opacity = self.getColorOpacityForVolumeRender()
 
         self.volume_property.SetColor(colors)
+
+        self.setDefaultScalarOpacity(self.volume_property.GetScalarOpacity())
+        self.setDefaultGradientOpacity(self.volume_property.GetGradientOpacity())
```
paskino, line 692: The names of these methods are confusing:

1. they have `default` but require a parameter
2. they are called independently of the mode of volume opacity render

Also the code in the lines below seems a bit muddled up. I think what should happen is:

1. set data for mapper
2. define which opacity method to use, with methods like `setVolumeOpacityToScalar` and `setVolumeOpacityToGradient` or with `setVolumeRenderOpacityMethod` (this will need to check if `self.volume_render_initialised == True` before calling `updateVolumePipeline` or better put the check in the latter method)
3. select colormap (notice that this is handled by an instance member
4. create `color` and `opacity`
5. finish to install the pipeline

Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -835,7 +853,16 @@ def updateVolumePipeline(self):
...
+
+            # Update whether we use our calculated opacity as the scalar or gradient opacity
+            # Also return the other opacity type to its default value:
+            if self.getVolumeRenderOpacityMethod() == 'gradient':
+                self.volume_property.SetScalarOpacity(self.getDefaultScalarOpacity()) 
+                self.volume_property.SetGradientOpacity(opacity)
+                 
+            elif self.getVolumeRenderOpacityMethod() == 'scalar': 
+                self.volume_property.SetGradientOpacity(self.getDefaultGradientOpacity())
+                self.volume_property.SetScalarOpacity(opacity)
```
paskino, line 865: You'll need to change these, reflecting the updates in the methods names above

paskino, line 865: You'll need to change these, reflecting the updates in the methods names above

paskino (changes requested):
I think you need to change some method names to make it clearer.

Wrappers/Python/ccpi/viewer/CILViewer.py
lauramurgatroyd: It seems the way it works (or what I understand) is that there is always a 'gradient opacity' and a 'scalar opacity'. which is why before we set any opacities we have to save the 2 default values, so they can be used later when we change the method we use. I could change the name instead to 'original' instead of default?
regarding point 2:
 - `updateVolumePipeline` already has the check for `volume_render_initialised`. 
 - I don't understand why we would use `setVolumeRenderOpacity` method at this point, as this will have been set by the user, or if it hasn't been set then when we later call `self.getVolumeRenderOpacityMethod()` we get 'gradient'

lauramurgatroyd: I'll check on the colormap code and make some additions there, I'll add changing the colormap to the example to help with that

Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -835,7 +853,16 @@ def updateVolumePipeline(self):
...
         
 
     def setVolumeColorLevelWindow(self, cmin, cmax):
```
lauramurgatroyd, line 868: Is the intention that this could be set by the user?
would it make more sense for the user to set the percentiles? I am not sure

lauramurgatroyd:
I updated the code so that when setting the gradient opacity we use the histogram statistics from the gradient of the image. But it is not looking great so far
egg with gradient opacity, unclipped and then clipped - most of it is at the high end of the colourmap
image
image

lauramurgatroyd:
Fixed issues above and added extra functionality. Updated PR description to reflect this.

lauramurgatroyd:
And this is what the egg looks like now - much better!

image

Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -712,52 +726,264 @@ def addHeadlight(self):
...
+        Returns
+        -----------
+        (min, max): tuple, default: (80., 99.)
+            the upper and lower image values that the 
+            opacity will be mapped to if setVolumeRenderOpacityMethod
+            has been set to 'scalar'.
+        '''
+        return self.scalar_opacity_limits
+
+    def setVolumeColorLevelWindow(self, min, max, update_pipeline=True):
```
paskino, line 863: maybe `setVolumeColorMap` would be more appropriate, or `setVolumeColorMapping`?

Wrappers/Python/ccpi/viewer/utils/colormaps.py
```diff
@@ -1108,7 +1108,7 @@ def relu(x, xmin, xmax, scaling=1):
         if val < xmin or val > xmax:
             out.append(0)
         else:
-            out.append((val - xmin) / dx)
+            out.append((val - xmin)*scaling / dx)
```
paskino, line 1111: just for clarity, could you change the position of scaling?

```python
out.append( scaling * ((val-xmin) / dx) )
```

Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -827,30 +1053,27 @@ def updatePipeline(self, resetcamera = False):
...
+            # Update whether we use our calculated opacity as the scalar or gradient opacity
+            if self.getVolumeRenderOpacityMethod() == 'gradient':
+                # Also return the scalar opacity to its default value:
+                self.volume_property.SetScalarOpacity(self._getDefaultScalarOpacityFunction()) 
+                self.volume_property.DisableGradientOpacityOff()
+                self.volume_property.SetGradientOpacity(opacity)
+                 
+            elif self.getVolumeRenderOpacityMethod() == 'scalar': 
+                self.volume_property.DisableGradientOpacityOn()
+                self.volume_property.SetScalarOpacity(opacity)
```
paskino, line 1070: Do I understand correctly that you don't have to set both scalar and gradient opacity as you found before?

paskino (changes requested):
a couple of minor changes.

Wrappers/Python/ccpi/viewer/utils/colormaps.py
lauramurgatroyd: done

Wrappers/Python/ccpi/viewer/CILViewer.py
lauramurgatroyd: If we want to use the scalar opacity, we don't have to use a gradient opacity - hence we disable it.
If we want to use a gradient opacity, there still exists a scalar opacity. Currently what happens is if we switch to gradient it uses the default scalar opacity from vtk. Maybe instead we should use *our* default scalar opacity, or whatever scalar opacity the user has set

Wrappers/Python/ccpi/viewer/CILViewer.py
lauramurgatroyd: renamed to setVolumeColorWindow

Wrappers/Python/ccpi/viewer/CILViewer.py
lauramurgatroyd: added comments to code to explain why we do this

lauramurgatroyd:
Added changes as discussed

paskino:
Ready to merge then!

paskino (approved):

PR #231, opened 2022-05-04
update volume render and clipping when new 3D image is input

### LINKED ISSUE #229: Updating input to 3D viewer doesn't update the volume render
Only the slice actor is updated.
We need to call installVolumeRenderActorPipeline() or update the pipeline

lauramurgatroyd:
Closes #229 

- Updates volume render when new image is input
- Also clears any clipping planes made and turns off the clipping widget


paskino (approved):
I trust that you have tested it, looks good to me.

PR #234, opened 2022-05-09
Improvements to HDF5 readers and writers

lauramurgatroyd:
- No need to specify VTK array name when writing to HDF5
- Use single instead of double underscores for variable names
- Allow setting both 4D index and 4D slice index
- Add checks for whether filename has been set before trying to read it
- Update unit tests to cover above changes

### DIFF
--- a/Wrappers/Python/ccpi/viewer/utils/hdf5_io.py
+++ b/Wrappers/Python/ccpi/viewer/utils/hdf5_io.py
@@ -3,12 +3,12 @@ from vtk.util.vtkAlgorithm import VTKPythonAlgorithmBase
 import h5py
 import numpy as np
 from vtk.numpy_interface import dataset_adapter as dsa
+from vtk.util import numpy_support
 
 # Methods for reading and writing HDF5 files:
 
 
-def write_image_data_to_hdf5(filename, data, dataset_name, attributes={},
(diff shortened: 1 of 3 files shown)


Wrappers/Python/ccpi/viewer/utils/hdf5_io.py
```diff
@@ -55,64 +57,106 @@ def __init__(self):
...
+        return self._FileName
 
     def SetDatasetName(self, lname):
-        if lname != self.__DatasetName:
+        if lname != self._DatasetName:
             self.Modified()
-            self.__DatasetName = lname
+            if self._FileName is not None:
+                with h5py.File(self._FileName, 'r') as f:
+                    if not (lname in f):
```
paskino, line 121: does this look in the `key`s?

paskino (approved):
It looks OK to me. I have not tested it, but I trust you

Wrappers/Python/ccpi/viewer/utils/hdf5_io.py
lauramurgatroyd: It looks at the keys but also works if it's a path such as entry1/tomo_entry/data etc
https://docs.h5py.org/en/stable/high/group.html#h5py.Group.__contains__

PR #239, opened 2022-05-16
Reset slice and volume defaults when a new image is input to the 3D viewer

### LINKED ISSUE #235: Updating input image should reset orientation of volume render
- [ ]  reset orientation of camera

Also:

- [ ] reset whether slice is visible
- [ ] reset whether volume is visible
- [ ] reset which direction we are slicing X/Y/Z
- [ ] delete all clippings and turn off clipping widget

### LINKED ISSUE #238: Clipping planes not correctly updated when new image is selected in a specific case
If you do:

1. load image 1
2. turn on volume
3. clip
4. turn off clipping
5. turn off volume
7. load image 2 instead of 1
8. turn on volume
9. turn on clipping

then the size of the clipping widget is incorrect.

If you skip step 5 and 8, the size of clipping widget is correct.

lauramurgatroyd:
Closes #235 
Closes #238

### DIFF
--- a/Wrappers/Python/ccpi/viewer/CILViewer.py
+++ b/Wrappers/Python/ccpi/viewer/CILViewer.py
@@ -261,40 +261,51 @@ class CILInteractorStyle(vtk.vtkInteractorStyleTrackballCamera):
                 viewer.getRenderer().Render()
             else:
                 # print ("handling c")
-                planew = vtk.vtkImplicitPlaneWidget2()
-                
-                rep = vtk.vtkImplicitPlaneRepresentation()
-                world_extent = self.GetImageWorldExtent()
-                extent = [0, world_extent[0], 0, world_extent[1], 0, world_extent[2]]
-                rep.SetWidgetBounds(*extent)
-                planew.SetInteractor(viewer.getInteractor())
-                planew.SetRepresentation(rep)
-
-                rep.SetNormalToCamera()
-                rep.SetOutlineTranslation(False) # this means user can't move bounding box
-
-                plane = vtk.vtkPlane()
-                # should be in the focal point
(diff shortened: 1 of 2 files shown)


Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -622,18 +633,41 @@ def startRenderLoop(self):
...
-                    self.planew.SetEnabled(False)
-                    self.volume.GetMapper().RemoveAllClippingPlanes()
-                    self.clipping_plane_initialised = False
-                self.installVolumeRenderActorPipeline()
-                self.volume.VisibilityOn()
-            else:
-                self.volume_render_initialised = False
+            if self.clipping_plane_initialised:
+                self.planew.SetEnabled(False)
+                self.volume.GetMapper().RemoveAllClippingPlanes()
```
paskino, line 642: These 2 lines seem to be repeated below. Are they necessary?

paskino (approved):
Looks good to me. Have a look at my comment

Wrappers/Python/ccpi/viewer/CILViewer.py
Pasarus: These first 2 lines seem to clear what is already there, then the next 2 seem to fix another issue with the camera position caused by the clipping plane.

lauramurgatroyd: They aren't necessary, and have now been deleted.

lauramurgatroyd: Creating the new clipping plane overrides the old one anyway so there is no point deleting them in these 2 lines.

PR #247, opened 2022-05-23
fix github action by installing lib_egl packages

lauramurgatroyd:


Pasarus (approved):
Achieves what it needs to, Nice Work :+1:

PR #248, opened 2022-05-24
Web application containing CILViewer

Pasarus:
You need a python environment with Trame, and Pywebvue installed, alongside the normal requirements for CILViewer. You need to install it will pip, and ensure it is not the v2 release candidates.

I have been running this in a conda environment python version 3.9, I recommend you do the same.

To run the `CILViewer/Wrappers/Python/ccpi/web_viewer/web_app.py` file. You need to pass files or a directory that contains files to load.

For example in the `CILViewer/Wrappers/Python/ccpi/web_viewer` directory, run:
`python web_app.py CILViewer/Wrappers/Python/ccpi/web_viewer/data`
This should launch the Trame/Web viewer application.

PR #249, opened 2022-05-25
Add slice histogram chart

Pasarus:
The best way to test this is by starting the web app using the instructions from this PR: #248

### DIFF
--- a/Wrappers/Python/ccpi/web_viewer/trame_viewer.py
+++ b/Wrappers/Python/ccpi/web_viewer/trame_viewer.py
@@ -83,10 +83,12 @@ class TrameViewer:
         self.slice_window_range_slider = None
         self.slice_window_slider = None
         self.slice_level_slider = None
+        self.show_slice_histogram_switch = None
         self.toggle_volume_visibility = None
         self.opacity_radio_buttons = None
         self.color_choice = None
         self.clipping_switch = None
+        self.clipping_removal_button = None
         self.color_slider = None
         self.windowing_range_slider = None
         self.reset_cam_button = None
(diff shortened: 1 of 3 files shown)


Wrappers/Python/ccpi/web_viewer/trame_viewer.py
```diff
@@ -0,0 +1,799 @@
...
+    def show_slice_histogram(self, show_histogram):
+        self.cil_viewer.updateSliceHistogram()
+        if show_histogram:
+            self.cil_viewer.histogramPlotActor.VisibilityOn()
+        else:
+            self.cil_viewer.histogramPlotActor.VisibilityOff()
+        self.html_view.update()
+
+    def change_slice_number(self, slice_number):
+        self.cil_viewer.updateSliceHistogram()
```
lauramurgatroyd, line 796: We should move this call to updateSliceHistogram into the CILViewer's updatePipeline() - so then it occurs within the viewer so if we made it as an interaction in the viewer (without trame) then it would be automatically updated when the slice number is changed still.

lauramurgatroyd (changes requested):
Histogram is looking good, just a couple of requests:
- see comment on moving some code into the viewer
- when you toggle the slice off, if the slice histogram was previously displayed I think it should be turned off with the slice. e.g. here I have turned off the slice but the slice histo remains and of course I can't get rid of it without turning the slice back on:
image

Pasarus:
All requested changes have been added to the CILViewer2D PR for simplicity as there is a significant refactor including this code in that PR.

lauramurgatroyd (approved):
I tested the histogram changes made in the other branch and I am happy they work 👍

PR #253, opened 2022-06-09
Add a CILViewer2D to the web app

Pasarus:
To be reviewed after: #249 and #255
Do not merge first

You need a python environment with Trame, and Pywebvue installed, alongside the normal requirements for CILViewer. You need to install it will pip, and ensure it is not the v2 release candidates.

I have been running this in a conda environment python version 3.9, I recommend you do the same.

To run the `CILViewer/Wrappers/Python/ccpi/web_viewer/web_app.py` file. You need to pass files or a directory that contains files to load. To run the 2D viewer instead of the 3D viewer pass either `-d` or `--2D`.

For example in the C`ILViewer/Wrappers/Python/ccpi/web_viewer directory`, run:
`python web_app.py --2D CILViewer/Wrappers/Python/ccpi/web_viewer/data`
This should launch the Trame/Web viewer application.


Wrappers/Python/ccpi/web_viewer/README.md
```diff
@@ -5,4 +5,6 @@ Follow these steps:
 - Create the conda environment:
     - `mamba env create -f dev-environment.yml`
 - With the activated conda environment you can now start the web application:
-    - `python web_app.py path/to/folder/of/data/to/use`
\ No newline at end of file
+    - `python web_app.py path/to/folder/of/data/to/use`
+- If you want to use the 2D viewer pass the 2D arg to the script i.e. --2D or -d:
```
lauramurgatroyd, line 9: you could add info on your entry point here

lauramurgatroyd: I found --2D had to go before the path to the data, please can you make that clear here

lauramurgatroyd (changes requested):
Hi, I am going to review twice - this review is based on me running the 2D viewer and what i found, the second review will be after I've looked at the code.

It is looking good but I found some bugs with the window/level...
- When I load the app, the window/level looks good and the slider looks correct
image

- When I press auto window-level this happens:
image

- After messing around with various (all of the) settings, if I click " reset defaults", it looks like this:
image
So again, different to the above 2.

I noticed the tracing is really laggy - is it similar for you?

I don't see an option to show the slice histogram - is that expected?

Do we have plans to add the line profile?

I tried to load a nexus file and got this error:

> ERROR:root:Exception raised
> ERROR:root:KeyError('opacity')
> ERROR:root:Traceback (most recent call last):
>   File "D:\lhe97136\Miniconda3\envs\cilviewer_webapp\lib\site-packages\wslink\backends\aiohttp\__init__.py", line 411, in onMessage
>     results = func(*args, **kwargs)
>   File "D:\lhe97136\Miniconda3\envs\cilviewer_webapp\lib\site-packages\pywebvue\protocol.py", line 275, in update_state
>     self.app.set(
>   File "D:\lhe97136\Miniconda3\envs\cilviewer_webapp\lib\site-packages\pywebvue\utils\__init__.py", line 230, in __exit__
>     fn(**self._app.state)
>   File "D:\lhe97136\Work\CCPi\CILViewer\Wrappers\Python\ccpi\web_viewer\web_app.py", line 113, in change_model
>     TRAME_VIEWER.load_file(kwargs['file_name'], windowing_method=kwargs['opacity'])
> KeyError: 'opacity'
>

Wrappers/Python/ccpi/viewer/CILViewer2D.py
```diff
@@ -2146,3 +2114,18 @@ def uninstallPipeline2(self):
         elif self.vis_mode == CILViewer2D.RECTILINEAR_WIPE:
             # rectilinear wipe visualises 2 images in the same pipeline
             pass
+
+    def getSliceMapWindow(self, percentiles):
+        ia = vtk.vtkImageHistogramStatistics()
```
lauramurgatroyd, line 2119: This is not giving the map window for the slice, it is giving it for the whole 3D dataset.

Pasarus: So this is just a bad name for the function, effectively it's doing the same thing as the Volume namesake in the none-2D viewer, it's been renamed to slice here because it's 2D alone. Perhaps I should rename it to something like `getAllSliceMapWindow`?

lauramurgatroyd: Currently in the 2D viewer we adjust the window/level based on the values in the current slice, not in the whole image. So auto window-level just takes into account the values on that slice, rather than in the whole volume. In the 3D viewer it's based on the whole volume.

We want the 2D and 3D viewer to perform in the same way for the slice, so the question is which do we want them both to do?
@paskino what do you think?
This can be resolved in a separate issue/PR though

Wrappers/Python/ccpi/viewer/CILViewer2D.py
```diff
@@ -2146,3 +2114,18 @@ def uninstallPipeline2(self):
...
+        ia.SetInputData(self.img3D)
+        ia.Update()
```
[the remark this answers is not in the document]
Pasarus: Removed!

Wrappers/Python/ccpi/web_viewer/trame_viewer2D.py
```diff
@@ -84,10 +86,16 @@ def construct_drawer_layout(self):
             vuetify.VDivider(), "Choose background color", self.background_choice,
             vuetify.VDivider(), self.slice_interaction_section,
             vuetify.VDivider(),
-            "Use Ctrl + Click on the slice, to show the ROI of the current slice, Click and drag to resize and repositon",
+            "Use Ctrl + Click on the slice, to show the ROI of the current slice, Click and drag to resize and repositon.\n"
+            "Move the ROI by using the middle mouse button",
```
lauramurgatroyd: Typo in "reposition"

Pasarus: Fixed!

Wrappers/Python/ccpi/viewer/CILViewer2D.py
Pasarus: It gives the raw value (867, 867)

Wrappers/Python/ccpi/web_viewer/trame_viewer.py
```diff
@@ -374,3 +374,20 @@ def create_drawer_ui_elements(self):
         raise NotImplementedError(
             "This function is not implemented in the base class, but you can expect an implementation in it's sub"
             " classes.")
+
+            self.cil_viewer.getRenderer().Render()
```
lauramurgatroyd, line 378: i don't think this should be here

Pasarus: Ya know I think the merge broke a couple things, let me clean this up!
```suggestion
```

Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -1153,7 +1154,7 @@ def setInterpolateOff(self):
 
     def setSliceColorWindowLevelPercentiles(self, min_percentage, max_percentage):
         min_val, max_val = self.getVolumeMapWindow((min_percentage, max_percentage), 'scalar')
-        self.setScalarOpacityWindow(min_val, max_val)
+        self.setSliceColorWindowLevel(min_val, max_val)
```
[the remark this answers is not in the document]
lauramurgatroyd: I suggest we rename setSliceColorWindowLevelPercentiles -> setSliceColorWindowPercentiles

We also need to fix the behaviour of the method - I will make a separate PR for it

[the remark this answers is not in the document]
Pasarus: Just to confirm this is not required to change in this PR, as it was not made or renamed in this PR, but a later one?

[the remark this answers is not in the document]
lauramurgatroyd: yes, will be fixed separately, issue is here https://github.com/vais-ral/CILViewer/issues/258

Wrappers/Python/ccpi/web_viewer/trame_viewer.py
```diff
@@ -420,84 +252,42 @@ def construct_slice_window_range_slider(self):
...
-        if self.cmax > 100:
-            # Use actual values
-            min_value = self.cmin
-            max_value = self.cmax
-            step = 1
-            self.windowing_slider_is_percentage = False
+    def update_slice_data(self):
+        if hasattr(self.cil_viewer, "getSliceMapWindow"):
+            self.cmin, self.cmax = self.cil_viewer.getSliceMapWindow((0., 100.))
+            self.slice_window_range_defaults = self.cil_viewer.getSliceMapWindow((60., 95.))
```
lauramurgatroyd, line 264: why is this 60 and 95?

Wrappers/Python/ccpi/web_viewer/trame_viewer.py
```diff
@@ -420,84 +252,42 @@ def construct_slice_window_range_slider(self):
...
-                                    disabled=self.disable_3d,
-                                    thumb_label=True,
-                                    style="max-width: 300px")
-
-    def update_windowing_defaults(self, method="scalar"):
-        self.cmin, self.cmax = self.cil_viewer.getVolumeMapWindow((0., 100.), method)
-        self.windowing_defaults = self.cil_viewer.getVolumeMapWindow((80., 99.), method)
-        self.slice_window_range_defaults = self.cil_viewer.getVolumeMapWindow((5., 95.), "scalar")
+            self.cmin, self.cmax = self.cil_viewer.getVolumeMapWindow((0., 100.), "scalar")
+            self.slice_window_range_defaults = self.cil_viewer.getVolumeMapWindow((5., 95.), "scalar")
```
lauramurgatroyd, line 267: The slice window range defaults should not come from the VolumeMapWindow

lauramurgatroyd: We should replace this whole if/else with:

```python
ia = vtk.vtkImageHistogramStatistics()
ia.SetInputData(self.cil_viewer.voi.GetOutput())
ia.SetAutoRangePercentiles(0, 100)
ia.Update()
self.cmin, self.cmax = ia.GetAutoRange()
ia.SetAutoRangePercentiles(5.0, 95.)
ia.Update()
self.slice_window_range_defaults = ia.GetAutoRange()
ia.Update()
```

What this does is it sets cmin and cmax to the min and max values for the SLICE. In the 2D viewer that is all you need. In the 3D viewer, these are different to the min and max value for the entire image volume.

To get the min and max value for the image volume you need:
`cmin, cmax = self.cil_viewer.getVolumeMapWindow((0., 100.), "scalar") `, like you have at line 266.

So what you will need for the 3D viewer is to store two different sets of cmin and cmax, one which is for the slice slider and one which is for the volume slider. And the slice slider range needs updating whenever the slice number is changed.

lauramurgatroyd: What I will do later is add the code for getting the cmin and cmax for the slice and getting the window range defaults for the slice as methods in both the 2D and 3D viewer (with the same names), so that later we can call these instead of you having to use all this code in the web app.

lauramurgatroyd: But these additions to the viewer will be done after this PR is merged, just to clarify.

lauramurgatroyd (changes requested):

Wrappers/Python/ccpi/web_viewer/trame_viewer.py
Pasarus: This is close to the default values that are being used by the CILViewer2D I can find no reason why they are, I got to these numbers through trial and error and comparison with the (60. 95.) and default from the 2D viewer.

lauramurgatroyd: That's weird, because the 2D viewer sets the percentiles here:
https://github.com/vais-ral/CILViewer/blob/47f45a970d6931d34173c5418737f765ab83658b/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L1625

Pasarus: Yeah I know, it's a very weird behaviour that I can't find the source for.

lauramurgatroyd: what happens when you set 5, 95?

lauramurgatroyd: problem could've been related to incorrect window definition in viewer?

Pasarus: The min value was 0. up until about 60% but the viewer was setting it to about 130-150. which is roughly what we get when we use 60% as the min value.

Wrappers/Python/ccpi/web_viewer/trame_viewer.py
```diff
@@ -664,62 +321,60 @@ def change_window_level_detail_sliders(self, show_detailed):
...
+            " classes.")
 
-            self.cil_viewer.remove_clipping_plane()
+    def create_drawer_ui_elements(self):
+        raise NotImplementedError(
+            "This function is not implemented in the base class, but you can expect an implementation in it's sub"
+            " classes.")
 
-            self.cil_viewer.getRenderer().Render()
             self.cil_viewer.updatePipeline()
```
lauramurgatroyd, line 378:
```suggestion
            self.cil_viewer.updatePipeline()
```

Wrappers/Python/ccpi/web_viewer/trame_viewer.py
lauramurgatroyd: Part of the problem is the getSliceMapWindow method isn't getting the window for the slice (as previously discussed). So 5, 95 will be correct when the other issues are fixed.

Wrappers/Python/ccpi/web_viewer/trame_viewer2D.py
```diff
@@ -0,0 +1,192 @@
...
+        if not interpolation:
+            self.cil_viewer.imageSlice.GetProperty().SetInterpolationTypeToNearest()
+        else:
+            self.cil_viewer.imageSlice.GetProperty().SetInterpolationTypeToLinear()
+        self.cil_viewer.updatePipeline()
+
+    def auto_window_level(self):
+        cmin, cmax = self.cil_viewer.ia.GetAutoRange()
+        level = (cmin + cmax) / 2
+        window = (cmax - cmin) / 2
```
lauramurgatroyd, line 158:
```suggestion
        window = cmax - cmin
```

lauramurgatroyd:
We still have some small issues with window/level and some things we need to discuss re: this. I will make issues to cover these.

Otherwise, all working and great work 👍 :)

PR #254, opened 2022-06-10
Add an entrypoint for the web_app

Pasarus:
Adds an entry point so we can easily start the web app from the command-line after pip installing the CILViewer to do this call:
`cilwebapp`


lauramurgatroyd (approved):
I tested (pip installing) and am happy it works

PR #268, opened 2022-06-20
Rename "Window" to "Range" where appropriate

### LINKED ISSUE #267: Some methods in viewer are called "window" when they should be called "range"
e.g.:
https://github.com/vais-ral/CILViewer/blob/86b1c569957a30f20f46fb9707b134c1320c6cd1/Wrappers/Python/ccpi/viewer/CILViewer.py#L872

This sets the range of values - i,e, min and max.

The window is a single value =max - min.

Therefore I suggest we rename to setGradientOpacityRange - and similar for other methods in this file.

lauramurgatroyd:
closes #267

### DIFF
--- a/Wrappers/Python/ccpi/web_viewer/dev-environment.yml
+++ b/Wrappers/Python/ccpi/web_viewer/dev-environment.yml
@@ -1,5 +1,5 @@
 # run: conda env create --file environment.yml
-name: cilviewer_webapp_trame2
+name: cilviewer_webapp
 dependencies:
 - python==3.9
 - matplotlib # Optional for more colormaps
--- a/Wrappers/Python/ccpi/web_viewer/dev-environment.yml
+++ b/Wrappers/Python/ccpi/web_viewer/dev-environment.yml
@@ -7,5 +7,5 @@ dependencies:
 - pip
 - pip:
   # Have to install Trame via pip due to unavailability on conda
-  - trame  # pin version for pip
+  - trame<2.0  # pin version for pip
   - vtk==9.1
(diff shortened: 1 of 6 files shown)


lauramurgatroyd:
I only thought of this afterwards but if you prefer we could use "Limits" instead of "Range" ?

Wrappers/Python/ccpi/web_viewer/dev-environment.yml
```diff
@@ -1,11 +1,11 @@
 # run: conda env create --file environment.yml
-name: cilviewer_webapp
+name: cilviewer_webapp_trame2
```
Pasarus, line 2: I don't like this renaming it adds unnecessary length to the name that you'd have to type.

Wrappers/Python/ccpi/web_viewer/dev-environment.yml
```diff
@@ -1,11 +1,11 @@
...
 dependencies:
 - python==3.9
 - matplotlib # Optional for more colormaps
 - h5py
 - pip
 - pip:
   # Have to install Trame via pip due to unavailability on conda
-  - trame<2.0  # pin version for pip
+  - trame  # pin version for pip
```
Pasarus, line 10: Should be handled in the other PR https://github.com/vais-ral/CILViewer/pull/262

Pasarus (changes requested):
A few comments before I look into the code functionality

lauramurgatroyd:
Sorry Sam I didn't mean to push my changes to the environment file! They were from when i was trying to get trame2 working!

Wrappers/Python/ccpi/web_viewer/dev-environment.yml
lauramurgatroyd, line 2:
```suggestion
name: cilviewer_webapp
```

Wrappers/Python/ccpi/web_viewer/dev-environment.yml
lauramurgatroyd, line 10:
```suggestion
  - trame<2.0  # pin version for pip
```

PR #269, opened 2022-06-21
Refactor to reduce complication

Pasarus:
An easy refactor to review, just code changes.

### DIFF
--- a/Wrappers/Python/ccpi/web_viewer/trame_viewer.py
+++ b/Wrappers/Python/ccpi/web_viewer/trame_viewer.py
@@ -368,11 +368,9 @@ class TrameViewer:
         self.html_view.update()
 
     def change_slice_number(self, slice_number):
-        if hasattr(self.cil_viewer, "updateSliceHistogram"):
-            self.cil_viewer.updateSliceHistogram()
-        self.cil_viewer.setActiveSlice(slice_number)
-        self.cil_viewer.updatePipeline()
-        self.html_view.update()
+        raise NotImplementedError(
+            "This function is not implemented in the base class, but you can expect an implementation in it's sub"
+            " classes.")
 
     def construct_drawer_layout(self):
         raise NotImplementedError(
--- a/Wrappers/Python/ccpi/web_viewer/trame_viewer.py
+++ b/Wrappers/Python/ccpi/web_viewer/trame_viewer.py
@@ -266,14 +266,9 @@ class TrameViewer:
         )
 
     def update_slice_data(self):
-        if hasattr(self.cil_viewer, "getSliceMapRange"):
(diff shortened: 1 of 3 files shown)


Wrappers/Python/ccpi/web_viewer/trame_viewer.py
```diff
@@ -329,50 +325,24 @@ def change_window_level_detail_sliders(self, show_detailed: bool):
...
+        raise NotImplementedError(
+            "This function is not implemented in the base class, but you can expect an implementation in it's sub"
+            " classes.")
 
     def change_slice_number(self, slice_number):
-        if hasattr(self.cil_viewer, "updateSliceHistogram"):
-            self.cil_viewer.updateSliceHistogram()
-        self.cil_viewer.setActiveSlice(slice_number)
-        self.cil_viewer.updatePipeline()
-        self.html_view.update()
```
lauramurgatroyd, line 375: you could still leave these here, then the 2D viewer would use the parent's change_slice_number unmodified, and the 3D viewer could run self.cil_viewer.updateSliceHistogram() and then call change_slice_number from the parent

lauramurgatroyd (changes requested):
One small suggestion. Otherwise, I don't mind the changes although I am in the middle of making a PR which reverts a lot of them because I am making it so that the 2D and 3D viewers have the same named methods for changing the slice's window and level.

BTW I didn't check it still runs, but I assume you did.

Wrappers/Python/ccpi/web_viewer/trame_viewer3D.py
```diff
@@ -458,3 +458,45 @@ def show_slice_histogram(self, show_histogram):
...
+
+    def change_slice_window_range(self, window: float, level: float):
+        if self.slice_window_slider_is_percentage:
+            min_percentage = level - window / 2
+            max_percentage = window + level - window / 2
+            self.cil_viewer.setSliceColorPercentiles(min_percentage, max_percentage)
+        else:
+            self.cil_viewer.setSliceColorWindowLevel(window, level)
+        self.cil_viewer.updatePipeline()
+        self.html_view.update()
```
lauramurgatroyd, line 496: why is this percentage stuff a thing only in the 3D case?

lauramurgatroyd: Oh I think the percentiles method only exists in the 3D viewer currently? So another thing for me to update

Pasarus:
Waiting to take stock in whether this is required anymore after this PR is finished/merged: https://github.com/vais-ral/CILViewer/pull/270

Pasarus:
This work will be replicated on top of #270 as part of work to create tests for the web app.

Issue #215, opened 2022-03-24
Create a base class for the CILViewer and CILViewer2D


lauramurgatroyd:
https://github.com/vais-ral/CILViewer/blob/73c65d6676998843e945ce933331930f1870ab78/Wrappers/Python/ccpi/viewer/CILViewer.py#L364

https://github.com/vais-ral/CILViewer/blob/73c65d6676998843e945ce933331930f1870ab78/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L1188

To do first:

- [ ] For methods in the viewers, should they all start with an upper or lowercase letter?

Common methods with same functionality:

- [ ] getRenderer
- [ ] GetSliceOrientation
- [ ]  g/G/s/SetActiveSlice (different case g and s in each viewer)
- [ ] g/GetRenderWindow
- [ ] getInteractor
- [ ] set/getColourWindow/Level
- [ ] saveRender

Common methods with different functionality:
- [ ] addActor - these should really have same functionality but that is a separate issue covered here: https://github.com/vais-ral/CILViewer/issues/92
- [ ]  startRenderLoop
- [ ] setInput3DData
- [ ] setInputData
- [ ] setInputAsNumpy - not sure why these differ
- [ ] installPipeline
- [ ] updatePipeline
- [ ] a/AdjustCamera

paskino:
notice that the 2D viewer could inherit from [`vtkImageViewer2`](https://vtk.org/doc/nightly/html/classvtkImageViewer2.html#details) class.

PR #270, opened 2022-06-22
Add a Base Class for CILViewer and CILViewer2D

### LINKED ISSUE #274: Linker Error due to GetSliceOrientation being renamed to getSliceOrientation only on CILViewer
A new error popped up in my [iDVC ](https://github.com/TomographicImaging/iDVC/tree/tiff_import)app with [`tiff_cropped_reader` ](https://github.com/vais-ral/CILViewer/tree/tiff_cropped_reader)branch.

```
File "C:\Apps\miniconda3\envs\idvc_dev8\lib\site-packages\ccpi\viewer\viewerLinker.py", line 311, in __call__
    if (self.sourceVtkViewer.getSliceOrientation() != self.targetVtkViewer.getSliceOrientation()):
AttributeError: 'CILViewer2D' object has no attribute 'getSliceOrientation'
'Linked2DInteractorStyle' object has no attribute 'getSliceOrientation'
```

`getSliceOrientation` is a method of `CILViewer` https://github.com/vais-ral/CILViewer/blob/6f698b786a412a5d29475f0cf70c05bf589117e4/Wrappers/Python/ccpi/viewer/CILViewer.py#L546-L547

but on the `CILViewer2D` it is still called 
https://github.com/vais-ral/CILViewer/blob/6f698b786a412a5d29475f0cf70c05bf589117e4/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L1741-L1742

lauramurgatroyd:
- Adds base class with current common methods between 3D and 2D viewers.
- Later should modify more methods so they are common between the viewers. Could also add methods with NotImplementedError in base class, that are defined in the subclasses.
- In a separate PR will add a base class for the interactor style.

Addresses part of: #215 
Closes #274

### DIFF
--- a/Wrappers/Python/ccpi/viewer/CILViewer.py
+++ b/Wrappers/Python/ccpi/viewer/CILViewer.py
@@ -16,12 +16,13 @@
 
 import numpy
 import vtk
-from ccpi.viewer.CILViewerBase import (ALT_KEY, CONTROL_KEY, CROSSHAIR_ACTOR, CURSOR_ACTOR, HELP_ACTOR, HISTOGRAM_ACTOR,
-                                       LINEPLOT_ACTOR, OVERLAY_ACTOR, SHIFT_KEY, SLICE_ACTOR, SLICE_ORIENTATION_XY,
-                                       SLICE_ORIENTATION_XZ, SLICE_ORIENTATION_YZ)
-from ccpi.viewer.utils import colormaps
-
+from ccpi.viewer  import (ALT_KEY, CONTROL_KEY, CROSSHAIR_ACTOR, CURSOR_ACTOR,
+                         HELP_ACTOR, HISTOGRAM_ACTOR, LINEPLOT_ACTOR,
+                         OVERLAY_ACTOR, SHIFT_KEY, SLICE_ACTOR,
+                         SLICE_ORIENTATION_XY, SLICE_ORIENTATION_XZ,
+                         SLICE_ORIENTATION_YZ)
 from ccpi.viewer.CILViewerBase import CILViewerBase
+from ccpi.viewer.utils import colormaps
 
 
 class CILInteractorStyle(vtk.vtkInteractorStyleTrackballCamera):
@@ -430,10 +431,7 @@ class CILViewer(CILViewerBase):
         CILViewerBase.__init__(self, dimx=600, dimy=600, ren=None, renWin=None, iren=None, debug=True)
         '''creates the rendering pipeline'''
 
-        # create a renderwindowinteractor
-        self.style = CILInteractorStyle(self)
-        self.iren.SetInteractorStyle(self.style)
-        self.iren.SetRenderWindow(self.renWin)
+        self.setInteractorStyle(CILInteractorStyle(self))
 
         self.sliceActorNo = 0
         # Render decimation
@@ -469,12 +467,6 @@ class CILViewer(CILViewerBase):
         self.volume_colormap_name = 'viridis'
         self.volume_render_initialised = False
         self.clipping_plane_initialised = False
-
-        self.iren.Initialize()
-
-    def getCamera(self):
-        '''returns the active camera'''
-        return self.ren.GetActiveCamera()
 
     def createPolyDataActor(self, polydata):
         '''returns an actor for a given polydata'''
--- a/Wrappers/Python/ccpi/viewer/CILViewer.py
+++ b/Wrappers/Python/ccpi/viewer/CILViewer.py
@@ -14,17 +14,15 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-import glob
-import os
-import re
-
 import numpy
 import vtk
-from ccpi.viewer.CILViewer2D import (ALT_KEY, CONTROL_KEY, CROSSHAIR_ACTOR, CURSOR_ACTOR, HELP_ACTOR, HISTOGRAM_ACTOR,
+from ccpi.viewer.CILViewerBase import (ALT_KEY, CONTROL_KEY, CROSSHAIR_ACTOR, CURSOR_ACTOR, HELP_ACTOR, HISTOGRAM_ACTOR,
                                      LINEPLOT_ACTOR, OVERLAY_ACTOR, SHIFT_KEY, SLICE_ACTOR, SLICE_ORIENTATION_XY,
-                                     SLICE_ORIENTATION_XZ, SLICE_ORIENTATION_YZ, ViewerEventManager)
+                                     SLICE_ORIENTATION_XZ, SLICE_ORIENTATION_YZ)
 from ccpi.viewer.utils import colormaps
-from ccpi.viewer.utils.io import SaveRenderToPNG
+
+
+from ccpi.viewer.CILViewerBase import CILViewerBase
 
 
 class CILInteractorStyle(vtk.vtkInteractorStyleTrackballCamera):
@@ -227,7 +225,7 @@ class CILInteractorStyle(vtk.vtkInteractorStyleTrackballCamera):
         # reset color/window
         cmin, cmax = self._viewer.ia.GetAutoRange()
 
-        window, level = self.getSliceWindowLevelFromRange(cmin, cmax)
+        window, level = self._viewer.getSliceWindowLevelFromRange(cmin, cmax)
 
         self.SetInitialLevel(level)
         self.SetInitialWindow(window)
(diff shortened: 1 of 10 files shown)


Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -526,50 +470,12 @@ def __init__(self, dimx=600, dimy=600, renWin=None, iren=None, ren=None, debug=F
...
-        '''returns the render window'''
-        return self.renWin
-
-    def getInteractor(self):
-        '''returns the render window interactor'''
-        return self.iren
-
     def getCamera(self):
         '''returns the active camera'''
         return self.ren.GetActiveCamera()
```
paskino, line 477: Why this is not in the base class?

Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -632,9 +538,6 @@ def addActor(self, actor):
         '''Adds an actor to the render'''
         return self.showActor(0, actor)
 
-    def startRenderLoop(self):
-        self.iren.Start()
-
     def setInput3DData(self, imageData):
         self.img3D = imageData
```
paskino, line 542: Also this one could be in the base class?

Wrappers/Python/ccpi/viewer/CILViewerBase.py
```diff
@@ -0,0 +1,278 @@
...
+ALT_KEY = -128
+
+SLICE_ACTOR = 'slice_actor'
+OVERLAY_ACTOR = 'overlay_actor'
+HISTOGRAM_ACTOR = 'histogram_actor'
+HELP_ACTOR = 'help_actor'
+CURSOR_ACTOR = 'cursor_actor'
+CROSSHAIR_ACTOR = 'crosshair_actor'
+LINEPLOT_ACTOR = 'lineplot_actor'
+WIPE_ACTOR = 'wipe_actor'
```
paskino, line 19: I guess these could be defined in `ccpi.viewer` rather than in `CILViewerBase`

Wrappers/Python/ccpi/viewer/CILViewer2D.py
```diff
@@ -1134,66 +1084,35 @@ def DisplayLineProfile(self, interactor, event, display):
...
-
-        self.renWin.SetSize(dimx, dimy)
-        self.renWin.AddRenderer(self.ren)
+        CILViewerBase.__init__(self, dimx=600, dimy=600, ren=None, renWin=None, iren=None, debug=True)
 
         self.style = CILInteractorStyle(self)
-        self.style.debug = debug
-
         self.iren.SetInteractorStyle(self.style)
         self.iren.SetRenderWindow(self.renWin)
```
paskino, line 1098: I guess this line needs to be documented in the base class, because it is necessary to set the render window to the interactor and this is not done in the base class

Wrappers/Python/ccpi/viewer/CILViewerBase.py
```diff
@@ -0,0 +1,278 @@
...
+    def isActive(self, event):
+        return self.events[event]
+
+    def isAllInactive(self):
+        """Returns True if all events are inactive"""
+        return all(not x for x in self.events.values())
+
+
+class CILViewerBase():
+
```
paskino, line 59: needs to document at least that the interactor needs to be told which render window it's meant to use. See comment above.

Wrappers/Python/ccpi/viewer/CILViewerBase.py
```diff
@@ -0,0 +1,278 @@
...
+    def getVolumeRange(self, method):
+        '''
+        Parameters
+        -----------
+        method: string : ['scalar', 'gradient']
+            'scalar' - returns full range of values in image
+            'gradient' - returns full range of values in image gradient
+        '''
+
+        return self.getVolumeMapRange((0, 100), method)
```
paskino, line 208: I understand why you'd get the min and max from 0 and 100 percentiles, but wouldn't it be clearer to do 
```python

ia = self.getImageHistogramStatistics(method)
return ia.GetMinimum(), ia.GetMaximum()
```

Wrappers/Python/ccpi/viewer/CILViewerBase.py
```diff
@@ -0,0 +1,278 @@
...
+        else:
+            grad = vtk.vtkImageGradientMagnitude()
+            grad.SetInputData(self.img3D)
+            grad.SetDimensionality(3)
+            grad.Update()
+            ia.SetInputData(grad.GetOutput())
+        ia.Update()
+        return ia
+
+    def getVolumeMapRange(self, percentiles, method):
```
paskino, line 187: I'm not a fan of the term `volume` but as long as it's documented that with `slice` we mean `a 2D slice of the 3D image` and with `volume` we mean `whole 3D image`, I'm fine with that

paskino (changes requested):
a couple of suggestions, especially regarding documenting behaviour

Pasarus (changes requested):
Should we name the base class CILViewer, with the 2D and 3D specialisations being CILViewer2D and CILViewer3D respectively? this is how I did it with the web app, regardless they should have the same naming style.

Could we also make changes to change_slice_window in the trame viewer on roughly line 341, the window range and level have had the support added for the base class but not the detailed window slider's function!

Wrappers/Python/ccpi/viewer/CILViewerBase.py
lauramurgatroyd: In the init file?

Wrappers/Python/ccpi/viewer/CILViewerBase.py
```diff
@@ -0,0 +1,278 @@
...
+        self.ren.Render()
+        self.renWin.Render()
+
+    def setSliceColorLevel(self, level):
+        self.imageSlice.GetProperty().SetColorLevel(level)
+        self.imageSlice.Update()
+        self.ren.Render()
+        self.renWin.Render()
+
+    def getSliceMapRange(self, percentiles):
```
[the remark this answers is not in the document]
lauramurgatroyd: Is there an aim for different functionality? I thought we wanted the same functionality in both?
We said that for the slice's window-level slider we can't update it every time based on the min/max values so in both cases shouldn't we be using the full volume range?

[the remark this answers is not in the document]
lauramurgatroyd: I am working at fixing regarding the above in a separate PR

Wrappers/Python/ccpi/viewer/CILViewerBase.py
lauramurgatroyd: What would you prefer instead of 'volume' to mean the whole image?

Wrappers/Python/ccpi/viewer/CILViewerBase.py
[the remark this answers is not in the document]
Pasarus: Yeah, so I have no idea what I was talking about yesterday. I was very wrong about my assumption here. The behaviour is essentially the same between the two.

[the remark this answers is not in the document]
Pasarus: I've moved the function for now here: https://github.com/vais-ral/CILViewer/commit/5b85b9d2c5b2f6a3407ea309f3f3e6cfe176e378#diff-de584c413901b0122c1c89ff8c8616cc2eb4add4616d167e15d0096c53992388=

Wrappers/Python/ccpi/viewer/CILViewer.py
lauramurgatroyd: It has different functionality in the 2 viewers as it involves resetting of clipping etc. in the 3D viewer

Wrappers/Python/ccpi/viewer/CILViewerBase.py
lauramurgatroyd: made a setInteractorStyle method in the base class

Wrappers/Python/ccpi/viewer/CILViewerBase.py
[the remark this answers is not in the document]
lauramurgatroyd: Okay, cool

Wrappers/Python/ccpi/viewer/CILViewerBase.py
lauramurgatroyd: I have moved to init file

Wrappers/Python/ccpi/viewer/CILViewer2D.py
lauramurgatroyd: moved SetRenderWindow line to base class
Documented setting of interactor style in base class

lauramurgatroyd:
I would prefer CILViewer2D and CILViewer3D also, but I don't have the time/patience right now to make sure nothing breaks when I change CILViewer->CILViewer3D.

Wrappers/Python/setup.py
```diff
@@ -36,7 +36,7 @@ def version2pep440(version):
     return v_pep440
 
 
-git_version_string = subprocess.check_output('git describe', shell=True).decode("utf-8").rstrip()[1:]
+git_version_string = "23"  #subprocess.check_output('git describe', shell=True).decode("utf-8").rstrip()[1:]
```
lauramurgatroyd, line 39:
```suggestion
git_version_string = subprocess.check_output('git describe', shell=True).decode("utf-8").rstrip()[1:]
```

Wrappers/Python/ccpi/viewer/CILViewer.py
paskino: Does it here? I think the method could at least be `NotImplemented` in the base class.

Wrappers/Python/ccpi/viewer/CILViewerBase.py
paskino: It's clearly confusing: VTK calls it image, and with volume, I often think of the volume render. That's why I'd stick to image for the whole dataset and slice for the 2D subset.

paskino:
We currently rename
https://github.com/vais-ral/CILViewer/blob/6f698b786a412a5d29475f0cf70c05bf589117e4/Wrappers/Python/ccpi/viewer/__init__.py#L1-L2

Let's discuss how to rename the whole package, because although I like vortex I suspect we'd better keep our brand.

Wrappers/Python/ccpi/viewer/CILViewer.py
lauramurgatroyd: It is

lauramurgatroyd:
Ok I'll leave any renaming for later then

lauramurgatroyd:
I think I have made the requested changes

Wrappers/Python/ccpi/viewer/CILViewerBase.py
lauramurgatroyd: I changed to 'image'

paskino (approved):

Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -430,10 +429,7 @@ def __init__(self, dimx=600, dimy=600, renWin=None, iren=None, ren=None, debug=F
         CILViewerBase.__init__(self, dimx=600, dimy=600, ren=None, renWin=None, iren=None, debug=True)
         '''creates the rendering pipeline'''
 
-        # create a renderwindowinteractor
-        self.style = CILInteractorStyle(self)
-        self.iren.SetInteractorStyle(self.style)
-        self.iren.SetRenderWindow(self.renWin)
+        self.setInteractorStyle(CILInteractorStyle(self))
```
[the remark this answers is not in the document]
lauramurgatroyd: Will make an issue when we have merged this

PR #278, opened 2022-06-29
Qt examples use head from example_data

paskino:
Uses `example_data` to load a dataset in the examples.

I think that example data could return a reader rather than a `vtkImageData`, what do you think?


paskino:
Changelog updated

PR #305, opened 2022-08-09
Fix build variants so that we get different versions for each python + VTK combination

lauramurgatroyd:
Previously the variants were not being built as we had the setting:

`noarch: python`

### DIFF
--- a/.github/workflows/conda_build_and_publish.yml
+++ b/.github/workflows/conda_build_and_publish.yml
@@ -26,4 +26,6 @@ jobs:
         publish: ${{ github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags') }}
         test_all: ${{(github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')) || (github.ref == 'refs/heads/master')}}
         additional_apt_packages: 'libegl1-mesa libegl1-mesa-dev'
+        convert_win: 'true'
+        convert_osx: 'true'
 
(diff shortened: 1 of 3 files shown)


Wrappers/Python/conda-recipe/meta.yaml
```diff
@@ -9,7 +9,7 @@ build:
   skip: True # [py==38 and np==115]
   preserve_egg_dir: False 
   number: {{ GIT_DESCRIBE_NUMBER }}
-  noarch: python
+  # noarch: generic
```
lauramurgatroyd, line 12:
```suggestion
```

lauramurgatroyd:
I tested locally and made sure it built all 5 versions as expected

.github/workflows/conda_build_and_publish.yml
```diff
@@ -26,4 +26,6 @@ jobs:
         publish: ${{ github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags') }}
         test_all: ${{(github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')) || (github.ref == 'refs/heads/master')}}
         additional_apt_packages: 'libegl1-mesa libegl1-mesa-dev'
+        convert_win: 'true'
+        convert_osx: 'true'
```
paskino, line 30: we don't want this, as the conversion made huge packages.

lauramurgatroyd: what do we do then? I need a windows version

paskino:
It looks like to me that `noarch` is correct. I acknowledge that in that configuration conda does not build the variants it should.

https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#architecture-independent-packages

Notice that with CIL we do build the packages for different architectures manually.

lauramurgatroyd:
No longer needed

PR #319, opened 2022-10-26
Method for deleting a widget from the viewer

### LINKED ISSUE #318: Add method for deleting a widget
We have methods for saving a reference to a widget and retrieving it:
https://github.com/vais-ral/CILViewer/blob/99194d0cb6b1b864c88852efe684eb10322b0429/Wrappers/Python/ccpi/viewer/CILViewerBase.py#L358-L363

It would be useful to have a method to delete a widget

lauramurgatroyd:
Closes #318

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 # Changelog
 
-## v22.4.0
+## vx.x.x
+- add `deleteWidget` method to CILViewerBase
 - Add environment file for development of the viewer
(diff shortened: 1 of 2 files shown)


paskino (approved):
it looks OK. Does `addWidgetReference` make sense as a method name?

lauramurgatroyd:
I don't like it either but I am not sure what to rename it to 🤔

Issue #323, opened 2022-11-07
ImageWriter doesn't write out with file extension set in FileFormat variable


lauramurgatroyd:
If the file format is one of these, it writes out as 'hdf5'. If we set the file format to 'nxs' for example, we should write out with extension 'nxs'.
https://github.com/vais-ral/CILViewer/blob/007e2ef2304f2fac270618bc7ebbca8e7c14fe7f/Wrappers/Python/ccpi/viewer/utils/io.py#L647-L649

paskino:
the fix should be quite easy?

```python
 if self._FileFormat in ['nxs', 'h5', 'hdf5', '']: 
     self._FileName = file_name + '.' + self._FileFormat
     writer = self._GetHDF5Writer() 
```

lauramurgatroyd:
yep, I had already made that fix locally, just hadn't made the PR yet

PR #326, opened 2022-11-08
ImageWriter writes out with file extension set in FileFormat variable

lauramurgatroyd:
Closes: https://github.com/vais-ral/CILViewer/issues/323


Wrappers/Python/ccpi/viewer/utils/io.py
```diff
@@ -645,7 +645,9 @@ def _GetWriter(self):
         file_name = os.path.splitext(self._FileName)[0]
 
         if self._FileFormat in ['nxs', 'h5', 'hdf5', '']:
-            self._FileName = file_name + '.hdf5'
+            if self._FileFormat == '':
+                self._FileFormat = 'h5'
```
paskino, line 649:
```suggestion
                self._FileFormat = 'hdf5'
```

PR #329, opened 2022-11-17
Move methods for resetting camera into the viewer

### LINKED ISSUE #327: Code for saving initial camera should be moved into the viewer
Currently this lives in the web viewer.
 
a CameraData object stores the original camera, so that it can be reset to later: https://github.com/vais-ral/CILViewer/blob/5953150362a94195d4284b0a933497b59f856919/Wrappers/Python/ccpi/web_viewer/trame_viewer3D.py#L100

https://github.com/vais-ral/CILViewer/blob/5953150362a94195d4284b0a933497b59f856919/Wrappers/Python/ccpi/web_viewer/trame_viewer3D.py#L372

lauramurgatroyd:
Closes #327 

Previously there was some code in the viewer for resetting to the original camera, but it did not work properly. Instead, the resetting was done properly in the web viewer. I moved the web viewer code into the main viewer.


Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -604,21 +609,16 @@ def installPipeline(self):
...
-        camera.SetPosition(self.getCamera().GetPosition())
-        camera.SetViewUp(self.getCamera().GetViewUp())
-        self.default_camera = camera
+        self.default_camera = CameraData(self.getCamera())
 
     def resetCameraToDefault(self):
         if hasattr(self, 'default_camera'):
-            self.ren.SetActiveCamera(self.default_camera)
+            self.adjustCamera(resetcamera=True)
+            self.default_camera.copy_data_to_other_camera(self.getCamera())
```
paskino, line 621: I'd drop `other` because there's no other camera in this call.
Also `copy_data` seems to be generic. I think this is also very confusing as the data that is copied is not quite explicit.

Why not making a `staticmethod` in `CameraData` along these lines

```python

class CameraData(object):

    @staticmethod
    def CopyDataToCamera(camera_data, vtkcamera):
        ...

```

This would then be used as 
```python

CameraData.CopyDataToCamera(self.default_camera, self.getCamera())
```

Wrappers/Python/ccpi/viewer/utils/camera_data.py
```diff
@@ -17,7 +17,8 @@
 #
 
 from dataclasses import dataclass
-from vtkmodules.vtkRenderingCore import vtkCamera
+
```
paskino, line 20: I'd rename the file to `CameraData`?

paskino (changes requested):
I don't understand why some unittests have been removed.

Wrappers/Python/ccpi/viewer/CILViewer.py
lauramurgatroyd: sounds good

lauramurgatroyd: I'll also rename self.default_camera to self.default_camera_data

Wrappers/Python/ccpi/web_viewer/test/test_trameviewer3d.py
```diff
@@ -483,15 +483,6 @@ def test_load_file_updates_windowing_defaults(self, _):
...
 
-    @mock.patch("ccpi.web_viewer.trame_viewer.TrameViewer.load_file")
-    def test_load_file_saves_a_new_camera_position(self, _):
-        cam_data = mock.MagicMock()
-        self.trame_viewer.original_cam_data = cam_data
-
-        self.trame_viewer.load_file("path")
-
-        self.assertNotEqual(self.trame_viewer.original_cam_data, cam_data)
-
```
lauramurgatroyd, line 494: Some of the unit tests have been removed because the variables they were referring to in the trame viewer no longer exist.

For example, this unit test got deleted because we no longer save the camera position in the trame viewer. But I could change it so that it tests the equivalent variables in the CILViewer.

Wrappers/Python/ccpi/web_viewer/dev-environment.yml
```diff
@@ -1,5 +1,5 @@
 # run: conda env create --file environment.yml
-name: cilviewer_webapp
+name: cilviewer_webapp4
```
lauramurgatroyd, line 2:
```suggestion
name: cilviewer_webapp
```

lauramurgatroyd:
Updated to include your suggestions and tidy up what was happening when a new image was loaded on the viewer.

Wrappers/Python/ccpi/viewer/CILViewer.py
```diff
@@ -604,21 +610,20 @@ def installPipeline(self):
...
+        self.default_camera_data = CameraData(self.getCamera())
 
     def resetCameraToDefault(self):
-        if hasattr(self, 'default_camera'):
-            self.ren.SetActiveCamera(self.default_camera)
+        ''' resets to the default camera settings for the current
+        loaded 3D image'''
+        if hasattr(self, 'default_camera_data'):
+            self.adjustCamera(resetcamera=True)
+            CameraData.CopyDataToCamera(self.default_camera_data, self.getCamera())
```
paskino, line 626: I'm not particularly fond of this method name, but I could live with it.

paskino (approved):

Wrappers/Python/ccpi/viewer/CILViewer.py
lauramurgatroyd: I have added docstrings

PR #337, opened 2023-02-20
Viewer toolbar

Dagonite:
Implementation of the viewer GUI toolbar.

### DIFF
--- a/Wrappers/Python/ccpi/viewer/ui/SettingsDialog.py
+++ b/Wrappers/Python/ccpi/viewer/ui/SettingsDialog.py
@@ -12,7 +12,6 @@ class SettingsDialog(FormDialog):
 
     def __init__(self, parent=None, title=None):
         FormDialog.__init__(self, parent, title=title)
-
         self.setWindowFlag(QtCore.Qt.WindowStaysOnTopHint, True)
         self.file_location = "."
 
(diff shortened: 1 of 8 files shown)


lauramurgatroyd:
Hi @Dagonite, thanks for opening the PR! We have a training course/hackathon coming up soon for our CIL software, so we are pretty busy over the next few weeks. This means we probably won't get on to looking at this yet, but we will get back to it after then!

Wrappers/Python/ccpi/viewer/ui/SettingsDialog.py
```diff
@@ -0,0 +1,158 @@
+import os
+
+from eqt.ui import FormDialog, UISliderWidget
+from PySide2 import QtCore, QtWidgets
+from vtkmodules.util import colors
```
lauramurgatroyd, line 5:
```suggestion
from vtk import vtkNamedColors as colors
```

Wrappers/Python/ccpi/viewer/ui/helpers.py
```diff
@@ -0,0 +1,47 @@
+from vtkmodules.util import colors
```
lauramurgatroyd, line 1:
```suggestion
from vtk import vtkNamedColors as colors
```

lauramurgatroyd (changes requested):
Unfortunately we cannot yet use imports from vtkmodules as we still have to support vtk8 for the iDVC app.

Instead we need to use: colors = vtk.vtkNamedColors()

This requires some more changes rather than just the import change so I will look into this another time

paskino:
Relevant changes are updated in #347

PR #362, opened 2023-07-31
Standalone viewer app defaults to downsample on Z

### LINKED ISSUE #359: Downsampling should use round rather than int
When calculating the downsampling on the Z axis we use `int`

https://github.com/vais-ral/CILViewer/blob/b6b0eb4ae2aa375bbaf5df8e11bd76cb1d739b28/Wrappers/Python/ccpi/viewer/utils/conversion.py#L1015-L1016

It should be using `round`

```python

num_slices_per_chunk = int(np.round(1 / xy_axes_magnification))  # number of slices in the z direction we are resampling together.
```

paskino:
The standalone app now defaults to downsample in all 3 dimensions, unless it's loading an HDF5 where it can infer that we are importing acquisition data.

closes #359


lauramurgatroyd:
Some of the unit tests are also using 'int' instead of 'round' which is why the gh-actions are failing.

paskino:
This PR fixes #359 by moving some functionality to `ccpi.viewer.utils.conversion` and uses `pytest` for testing.

Wrappers/Python/test/test_io.py
```diff
@@ -4,23 +4,9 @@
 import h5py
 import numpy as np
 import vtk
-from ccpi.viewer.utils.conversion import Converter
+from ccpi.viewer.utils.conversion import *
```
lauramurgatroyd, line 7: Have we actually agreed to move to 'import *'? It's not something we do in CIL and we haven't previously done it in the viewer?

Wrappers/Python/test/test_viewer_main_windows.py
```diff
@@ -31,30 +31,55 @@ class TestViewerMainWindow(unittest.TestCase):
...
-    def test_init_calls_super_init(self):
+    def _test_init_calls_super_init(self):
         vmw = ViewerMainWindow(title="Testing Title", app_name="testing app name")
         # If the super init is called, then the following attributes should be set:
         assert vmw.app_name == "testing app name"
         assert vmw.threadpool is not None
         assert vmw.progress_windows == {}
 
-    def test_createAppSettingsDialog_calls_setAppSettingsDialogWidgets(self):
+    def _test_createAppSettingsDialog_calls_setAppSettingsDialogWidgets(self):
```
lauramurgatroyd, line 74: What's the reason for the underscores in front of the methods?

paskino: https://github.com/vais-ral/CILViewer/issues/367 You can't run more than one QApplication so I thought to run the tests sequentially, which means not run them as indipendent tests.
Unfortunately this works to a certain extent, because you can't run 2 different test classes because they might run at the same time.

Wrappers/Python/test/test_io.py
paskino: I now removed * from the test files

lauramurgatroyd (commented):
I approve of the changes, except when I run locally I get the error:

```
_______________________________ TestViewerMainWindow.test_all ________________________________

self = <test.test_viewer_main_windows.TestViewerMainWindow testMethod=test_all>

    def test_all(self):
        global _instance
        if _instance is None:
>           _instance = QApplication(sys.argv)
E           RuntimeError: Please destroy the QApplication singleton before creating a new QApplication instance.

test_viewer_main_windows.py:42: RuntimeError

FAILED test_viewer_main_windows.py::TestViewerMainWindow::test_all - RuntimeError: Please destroy the QApplication singleton before creating a new QApplication...
```

paskino:
Possibly the failure is due to #367

PR #365, opened 2023-09-11
Viewer slider

paskino:
Adding a slider widget to control the slice that is being displayed. Both the mouse wheel and this slider can be used to change the slice.

A `SliceSliderRepresentation` with default colors (CIL pinks), orientation (`horizontal`) and position is introduced.

Closes #358 

The orientation and position of the slider can be changed.

I tested on Windows with the web app and it doesn't seem to be able to grab the slider.

image

image

PR #432, opened 2024-08-07
Hide slider when image dimension is 1

### LINKED ISSUE #421: Slider should disappear when only 1 slice is visualised
When only 1 slice is visualised, the slider shows values between -1 and 0. The slider should not appear and only one slice should be shown.

This is connected to #420

DanicaSTFC:
closes #421

### DIFF
--- a/Wrappers/Python/ccpi/viewer/CILViewer2D.py
+++ b/Wrappers/Python/ccpi/viewer/CILViewer2D.py
@@ -1592,10 +1592,9 @@ class CILViewer2D(CILViewerBase):
         # save references
         self.sliderWidget = sw
         self.sliderCallback = cb
-        
+
         # enable slider
         self.sliderWidget.SetEnabled(not slider_hidden)
-
 
     def uninstallSliderWidget(self):
         '''remove the slider widget from the viewer'''
--- a/Wrappers/Python/ccpi/viewer/CILViewer2D.py
+++ b/Wrappers/Python/ccpi/viewer/CILViewer2D.py
@@ -1550,11 +1550,18 @@ class CILViewer2D(CILViewerBase):
         '''Create the pipeline for the slice slider widget
         
         The slider widget and representation are created if not already present.
-        Currently the slider widget enabled flag is not used.
+        If present, the slider is updated.
+        The slider is hidden if any of the dimensions of the visualised image is 1,
+        else it is shown.
         '''
+        # set slider_hidden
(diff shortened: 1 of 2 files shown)


Wrappers/Python/ccpi/viewer/CILViewer2D.py
```diff
@@ -1586,6 +1592,9 @@ def installSliceSliderWidgetPipeline(self):
         self.sliderWidget = sw
         self.sliderCallback = cb
 
+        # enable slider
+        self.sliderWidget.SetEnabled(not slider_hidden)
```
paskino, line 1596: Why not put this at line 1576?

paskino:
From email from @DanicaSTFC 
> I know you have a flag for this already but I preferr to keep the hide/show flag separate from the creation flag.

At the time I did not find a way to hide the slider widget, which you seem to have found. I used `EnabledOn`.

https://github.com/TomographicImaging/CILViewer/blob/cf6cb86489d1962caa8e2195d9499c3ffbc6e2a7/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L1565-L1569

It would be useful to add a method `setSliderWidgetEnabled` to be able to control the visibility of the slider widget.

Wrappers/Python/ccpi/viewer/CILViewer2D.py
DanicaSTFC: I moved it

DanicaSTFC:
> This PR seems to fix the last point at [#365 (comment)](https://github.com/TomographicImaging/CILViewer/pull/365#issuecomment-1737227437)
> 
> From email from @DanicaSTFC
> 
> > I know you have a flag for this already but I preferr to keep the hide/show flag separate from the creation flag.
> 
> At the time I did not find a way to hide the slider widget, which you seem to have found. I used `EnabledOn`.
> 
> https://github.com/TomographicImaging/CILViewer/blob/cf6cb86489d1962caa8e2195d9499c3ffbc6e2a7/Wrappers/Python/ccpi/viewer/CILViewer2D.py#L1565-L1569
> 
> It would be useful to add a method `setSliderWidgetEnabled` to be able to control the visibility of the slider widget. This would allow a proper treatment of the visibility and creation of the widget.

I did not create a new method to control the visibility as the `SetEnabled` method already exists. The creation of the widget in a viewer is a different matter. I was not sure what you wanted to do with the enabled parameter. Perhaps we can discuss this in a meeting.

